PR #89, opened 2016-09-17
Allowing chesster to report joins

lakinwecker:


heltour/tournament/api.py
```diff
@@ -401,3 +401,24 @@ def get_private_url(request):
...
+def player_joined_slack(request):
+    try:
+        name = request.POST.get('name', None)
+    except ValueError:
+        return HttpResponse('Bad request', status=400)
+
+    if not name:
+        return HttpResponse('Bad request', status=400)
+    try:
+        player = Player.objects.get(lichess_username__iexact=name.lower())
```
cyanfish: .lower() is unnecessary here

lakinwecker: Good catch. I just made  PR that fixes this.

PR #101, opened 2016-09-25
Sending a ping every 2s actually keeps the connection alive so it works.

lakinwecker:
This is inspite of the bug filed for lila.


ornicar:
Please revert this. It's no longer necessary, I fixed the issue at lichess level.

PR #135, opened 2017-01-08
RFC - I started down the road of adding a blitz league.

lakinwecker:
The red theme is the green theme with the R and G portions flipped. :P  We should probably do a better job of that.

It works fairly well, there is one thing that I haven't figured out how I want to do yet. The site still shows the classical ratings of players instead of the blitz ratings. Thoughts?


lakinwecker:
Not sure if RFC is the right term here, but I want comments before I continue.

cyanfish:
For the theme: Code looks good. Primary color looks good. I'd tweak the secondary color a bit, it looks a bit too purple. I can do that if you want.

For the registration form: Looks fine for now. Going forward, I want to generalize it as much as possible (e.g. change the "Can commit" text based on `league.time_control` and `season.round_duration`), but that would depend on model changes that aren't in place yet.

For ratings: That's going to be the hardest change. The site was not designed with multiple ratings in mind.
- In terms of the model, we should probably add `league.rating_type`, `player.blitz_rating`, and `player.blitz_games_played` fields.
- Populating that data from the lichess api is easy enough, just need to add those fields to the existing code.
- We could use helper methods like `player.get_rating(league)` and similar template tags to get the correct rating in various parts of the code.
- It's going to touch a lot of the code, so I can definitely help with this.

PR #143, opened 2017-01-21
service files

lakinwecker:


niklasf:
Once the immediate problems are solved: There's also `User=...` instead of using `sudo -u`.

PR #147, opened 2017-01-24
League-specific permissions (for review)

cyanfish:


heltour/tournament/admin.py
```diff
@@ -45,19 +47,129 @@ class _BaseAdmin(VersionAdmin):
     change_form_template = 'tournament/admin/change_form_with_comments.html'
     history_latest_first = True
 
+    league_id_field = None
+    league_competitor_type = None
+    allow_all_staff = False
+
+    def has_assigned_perm(self, user, perm_type):
+        return 'tournament.%s_%s' % (perm_type, self.opts.model_name) in user.get_all_permissions()
```
lakinwecker: user.has_perm() didn't work here?

heltour/tournament/admin.py
```diff
@@ -45,19 +47,129 @@ class _BaseAdmin(VersionAdmin):
...
+
+    def has_league_perm(self, user, action, obj):
+        if self.league_id_field is None:
+            return False
+        authorized_leagues = self.authorized_leagues(user)
+        if self.league_competitor_type is not None \
+                and all((League.objects.get(pk=pk).competitor_type != self.league_competitor_type for pk in authorized_leagues)):
+            return False
+        if obj is None:
+            return len(authorized_leagues) > 0
```
lakinwecker: This can be done with: return bool(authorized_leagues) not sure which reads better. I'm fine with either

heltour/tournament/admin.py
```diff
@@ -45,19 +47,129 @@ class _BaseAdmin(VersionAdmin):
...
+
+        def clean(form):
+            super(ModelForm, form).clean()
+            self.clean_form(request, form)
+
+        form.clean = clean
+        return form
+
+    def clean_form(self, request, form):
+        if self.allow_all_staff:
```
lakinwecker: Should this also check `request.user.is_staff`?

heltour/tournament/admin.py
cyanfish: `has_perm` doesn't work exactly as it should because of the bug workaround in `auth.py` (see the comments in that file)

heltour/tournament/admin.py
```diff
@@ -45,19 +47,129 @@ class _BaseAdmin(VersionAdmin):
...
+        if self.league_competitor_type is not None \
+                and all((League.objects.get(pk=pk).competitor_type != self.league_competitor_type for pk in authorized_leagues)):
+            return False
+        if obj is None:
+            return len(authorized_leagues) > 0
+        else:
+            return getnestedattr(obj, self.league_id_field) in authorized_leagues
+
+    def get_queryset(self, request):
+        result = super(_BaseAdmin, self).get_queryset(request)
```
lakinwecker: I would slightly prefer calling `result` `queryset` instead

heltour/tournament/admin.py
```diff
@@ -45,19 +47,129 @@ class _BaseAdmin(VersionAdmin):
...
+                if parts[0][-3:] != '_id':
+                    raise ValueError('Invalid league id field on modeladmin')
+                league_id = form.cleaned_data[parts[0][:-3]].id
+        else:
+            league_id = getnestedattr(form.cleaned_data[parts[0]], parts[1])
+        if league_id not in self.authorized_leagues(request.user):
+            raise ValidationError('No permission to save objects for this league')
+
+    def authorized_leagues(self, user):
+        return [lm.league_id for lm in LeagueModerator.objects.filter(player__lichess_username__iexact=user.username)]
```
lakinwecker: You should use the `.values` query here if you're only getting IDs. It will be a fair bit faster

heltour/tournament/admin.py
```diff
@@ -45,19 +47,129 @@ class _BaseAdmin(VersionAdmin):
...
+
+#-------------------------------------------------------------------------------
+class LeagueRestrictedListFilter(RelatedFieldListFilter):
+
+    def __init__(self, field, request, params, model, model_admin, field_path):
```
lakinwecker: This method doesn't do anything, we should remove it.

heltour/tournament/admin.py
```diff
@@ -157,42 +273,45 @@ class SeasonAdmin(_BaseAdmin):
...
             url(r'^(?P<object_id>[0-9]+)/mod_report/$',
-                permission_required('tournament.change_season')(self.admin_site.admin_view(self.mod_report_view)),
+                self.admin_site.admin_view(self.mod_report_view),
                 name='mod_report'),
         ]
         return my_urls + urls
 
 
     def simulate_tournament(self, request, queryset):
+        if not request.user.is_superuser:
```
lakinwecker: I would generally prefer specific permission checks over explicit is_superuser checks, but this shouldn't stop you from using this code.

cyanfish:
Will address the issues noted.

heltour/tournament/admin.py
cyanfish: I don't think non-staff can access the admin site at all, so it should be fine

heltour/tournament/admin.py
cyanfish: I'm probably going to remove this code entirely, it's more appropriate for a separate script

PR #173, opened 2017-05-29
Updated export_players information export

michael1241:
Added 'date_created' and 'friends' output in the export_player page.

str() called on date_created to make it JSON serialisable.

### DIFF
--- a/heltour/tournament/admin.py
+++ b/heltour/tournament/admin.py
@@ -675,7 +675,9 @@ class SeasonAdmin(_BaseAdmin):
                 'rating': sp.player.rating_for(season.league),
                 'has_20_games': sp.player.games_played_for(season.league) >= 20,
                 'in_slack': sp.player.in_slack_group,
-                'account_status': sp.player.account_status
+                'account_status': sp.player.account_status,
+		'date_created': sp.registration.friends if sp.registration else None,
+		'friends': (sp.registration.date_created if sp.registration else sp.date_created).isoformat()
             }
             reg = sp.registration
             if reg is not None:
--- a/heltour/tournament/admin.py
+++ b/heltour/tournament/admin.py
@@ -676,8 +676,8 @@ class SeasonAdmin(_BaseAdmin):
                 'has_20_games': sp.player.games_played_for(season.league) >= 20,
                 'in_slack': sp.player.in_slack_group,
                 'account_status': sp.player.account_status,
-                'date_created': sp.registration.friends if sp.registration else None,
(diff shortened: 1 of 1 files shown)


heltour/tournament/admin.py
```diff
@@ -661,7 +661,9 @@ def player_info_view(self, request, object_id, player_name):
             'league': season.league,
             'player': season_player.player,
             'reg': reg,
-            'has_played_20_games': has_played_20_games
+            'has_played_20_games': has_played_20_games,
+            'date_created': str(sp.registration.date_created or sp.player.date_created),
+            'friends': sp.registration.friends or None
```
lakinwecker: if registration is None both of these lines will fail. I would suggest using this format:

`sp.registration.friends if sp.registration else None`
and
`(sp.registration.date_created if sp.registration else sp.date_created).isoformat()`\

This also uses the `isoformat()`

lakinwecker: This also uses the date_created on the seasonplayer directly, not on the Player record which might be quite old.

heltour/tournament/admin.py
```diff
@@ -661,7 +661,9 @@ def player_info_view(self, request, object_id, player_name):
             'league': season.league,
             'player': season_player.player,
             'reg': reg,
-            'has_played_20_games': has_played_20_games
+            'has_played_20_games': has_played_20_games,
+            'date_created': sp.registration.friends if sp.registration else None,
+	    'friends': (sp.registration.date_created if sp.registration else sp.date_created).isoformat()
```
lakinwecker: Looks better, but I think there is a tab mixed in with the spaces here.  Can you verify that you're using spaces only for indentation? 4 spaces per level of indentation is what is proposed by PEP-8 (the python formatting guidelines) and is what we use in this project.

heltour/tournament/admin.py
```diff
@@ -661,7 +661,9 @@ def player_info_view(self, request, object_id, player_name):
             'league': season.league,
             'player': season_player.player,
             'reg': reg,
-            'has_played_20_games': has_played_20_games
+            'has_played_20_games': has_played_20_games,
+            'date_created': sp.registration.friends if sp.registration else None,
```
cyanfish: `sp` is not defined here, you'll either need to expand it to `season_player` or rename the variable above

cyanfish: Also, I think this is the wrong spot to add this. You probably want line 610, in which case `sp` would be defined.

PR #189, opened 2018-04-07
Added requests for players to avoid in team formation.

michael1241:
_crosses fingers_


lakinwecker:
https://github.com/cyanfish/heltour/blob/master/heltour/tournament/models.py#L1470
You will also need to add the field to the model and add the appropriate migration.

lakinwecker:
Ah yes. You did add it to the models, but you'll need a migration.

PR #204, opened 2018-06-10
Adding the chess960 league.

lakinwecker:
This is my initial changes to support the 960 league.  Let me know feedback, or if I missed anything. I'll work on chesster next.


heltour/tournament/models.py
```diff
@@ -77,9 +77,11 @@ class Meta:
     ('blue', 'Blue'),
     ('green', 'Green'),
     ('red', 'Red'),
+    ('yellow', 'Yellow'),
 )
 RATING_TYPE_OPTIONS = (
     ('classical', 'Classical'),
+    ('rapid', 'Rapid'),
```
cyanfish: Should also add "chess960" since this is the `perf` value, not the `speed` value.

cyanfish:
https://github.com/cyanfish/heltour/blob/4d5dc7033ed0ac3d1ed619e4fda5690db5e720bc/heltour/tournament/models.py#L679

`rating_for` and `games_played_for` will need to be updated too. I don't think having blitz-specific properties is needed, you should be able to remove those and generalize it.

Also, the registration form will need one description updated:
https://github.com/cyanfish/heltour/blob/28a3d0603a50ee735a440cd0deaf274b5119ac19/heltour/tournament/forms.py#L60

Looks good otherwise.

lakinwecker:
* I have fixed the form message.
  * I have fixed some theme issues.
  * I have made the rating methods more generic.

heltour/tournament/models.py
```diff
@@ -677,13 +662,18 @@ def is_available_for(self, round_):
...
+            return self.profile['perfs'].get(league.rating_type).get('rating')
+        return rating
 
     def games_played_for(self, league):
-        if league is not None and league.rating_type == 'blitz':
-            return self.blitz_games_played
+        if league:
+            if self.profile is None:
+                return None
+            return self.profile['perfs'].get(league.rating_type).get('games')
```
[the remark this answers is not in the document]
lakinwecker: Also fixed.

heltour/tournament/models.py
lakinwecker: I added chess960 here, but as per the convo in the dev channel, we'll be using Rapid for this as the lack of time-control specific ratings really messes with it, so the chess960 mods feel that Rapid is a better indication of playing strength than the chess960 rating.

heltour/tournament/models.py
```diff
@@ -677,13 +662,18 @@ def is_available_for(self, round_):
...
 
     def rating_for(self, league):
-        if league is not None and league.rating_type == 'blitz':
-            return self.blitz_rating
-        return self.rating # classical
+        if league:
+            if self.profile is None:
+                return None
+            return self.profile['perfs'].get(league.rating_type, {}).get('rating')
+        return rating
```
cyanfish: `self.rating`

heltour/tournament/templates/tournament/lone_league_home.html
```diff
@@ -64,7 +64,7 @@ <h3>{{ league.name }} TV</h3>
 		                		</a>
 	                		</div>
 		                </div>
-		                <div class="chessground small {% if league.theme == 'green' %}green{% else %}blue{% endif %} cburnett">
+                        <div class="chessground small {% if league.theme == 'green' %}green{% elif league.theme == 'blue' %}blue{% else %}yellow{% endif %} cburnett">
```
cyanfish: This class references
https://github.com/cyanfish/heltour/blob/master/heltour/tournament/static/lib/css/cg_theme.css

You could either add a new `yellow` class with a new bg image, or keep it simple and use the `green` class for now (or another one that already exists if you prefer).

heltour/tournament/templates/tournament/team_league_home.html
```diff
@@ -69,7 +69,7 @@ <h3>{{ league.name }} TV</h3>
 		                		</a>
 	                		</div>
 		                </div>
-		                <div class="chessground small {% if league.theme == 'green' %}green{% else %}blue{% endif %} cburnett">
+                        <div class="chessground small {% if league.theme == 'green' %}green{% elif league.theme == 'blue' %}blue{% else %}yellow{% endif %} cburnett">
```
cyanfish: Same as above

Issue #222, opened 2018-07-03
Career records for a player


aliquantus:
From season 12 survey: "Show career records for a player (in addition to season-specific records)"

PR #231, opened 2018-07-15
Career records

michael1241:
Show career record on the player profile page for the currently shown league type.

### DIFF
--- a/heltour/tournament/views.py
+++ b/heltour/tournament/views.py
@@ -1396,36 +1396,10 @@ class PlayerProfileView(LeagueView):
                 print(league_perf_obj._game_count)
             return league_perf_obj
 
-        career_performance_4545 = "N/A"
-        career_score_4545 = "N/A"
-        career_score_total_4545 = "N/A"
-        career_performance_LoneWolf = "N/A"
-        career_score_LoneWolf = "N/A"
-        career_score_total_LoneWolf = "N/A"
-        career_performance_BlitzBattle = "N/A"
-        career_score_BlitzBattle = "N/A"
-        career_score_total_BlitzBattle = "N/A"
-
         for leaguetype in other_season_leagues:
-            if leaguetype[0].name == "Lichess4545 League":
-                career_4545 = league_performance(leaguetype[1])
-                career_performance_4545 = career_4545.calculate()
-                career_score_4545 = career_4545._score
-                print(career_score_4545)
-                career_score_total_4545 = career_4545._game_count
-                print(career_score_total_4545)
-
-            if leaguetype[0].name == "Lichess LoneWolf":
-                career_LoneWolf = league_performance(leaguetype[1])
-                career_performance_LoneWolf = career_LoneWolf.calculate()
-                career_score_LoneWolf = career_LoneWolf._score
-                career_score_total_LoneWolf = career_LoneWolf._game_count
-
-            if leaguetype[0].name == "Blitz Battle":
-                career_BlitzBattle = league_performance(leaguetype[1])
-                career_performance_BlitzBattle = career_BlitzBattle.calculate()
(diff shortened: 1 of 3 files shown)


aliquantus:
Resolves issue #222

heltour/tournament/views.py
```diff
@@ -1345,6 +1345,59 @@ def team(season):
...
+                    season_score_total += 1
+            return season_perf
+
+        def league_performance(seasons):
+            league_perf_obj = PerfRatingCalc()
+            for season in seasons:
+                league_perf_obj.merge(season_performance(season[0]))
+            return league_perf_obj
+
+        for leaguetype in other_season_leagues:
```
cyanfish: In general, prefer deconstruction syntax rather than indexing into a tuple. For example,
`for league, season_info in other_season_leagues`

In this case, since you're not using most of that (convoluted) type anyway, you should just query for the data you want:
`[sp.season for sp in player.seasonplayer_set.filter(season__league=self.league)]`
And then you can get rid of this loop entirely, and make it clearer what data you are using in `league_performance`.

heltour/tournament/views.py
```diff
@@ -1345,6 +1345,59 @@ def team(season):
...
+            if leaguetype[0].tag == self.league.tag:
+                career = league_performance(leaguetype[1])
+                performances = [career.calculate(), career._score, career._game_count]
```
cyanfish: Avoid accessing private (`_`-prefixed) members. In this case, you probably want the overall game count rather than only games eligible for performance rating. Those are the `season_score` and `season_score_total` variables in the season_performance function.

If you like, you can expand the PerfRatingCalc class to include this data. For example add an optional parameter,
`add_game(self, score, opponent_rating, include_in_perf=True):`
And then a couple extra fields (public, or accessible via a `@property`) for `score` and `score_total`.

Also, you probably meant to create a tuple (with `()` brackets) instead of a list (with `[]`).

heltour/tournament/views.py
```diff
@@ -1345,6 +1345,59 @@ def team(season):
...
+                            if sp is not None and sp.seed_rating is not None:
+                                opp_rating = sp.seed_rating
+                            else:
+                                opp_rating = p.black_rating_display() if p.white == player else p.white_rating_display(self.league)
+                            season_perf.add_game(game_score, opp_rating)
+                elif round_.number in byes:
+                    bye = byes[round_.number]
+                    season_score += bye.score()
+                    season_score_total += 1
+            return season_perf
```
cyanfish: This should probably return `season_score` and `season_score_total` as well.

heltour/tournament/views.py
```diff
@@ -1345,6 +1345,59 @@ def team(season):
                 season_score += bye.score()
                 season_score_total += 1
         season_perf_rating = season_perf.calculate()
+        
+        def season_performance(season):
```
cyanfish: Most of the above code (the current season perf calculation) should be replaced with a call to this function (to avoid duplication).

heltour/tournament/views.py
```diff
@@ -1397,6 +1450,9 @@ def team(season):
             'season_perf_rating': season_perf_rating,
             'season_score': season_score,
             'season_score_total': season_score_total,
+            'career_performance': performances[0],
```
cyanfish: Again, prefer deconstruction (or just use separate variables).
If you want compactness, you can assign multiple related variables on the same line. For example,
`a, b, c = d, e, f`.

heltour/tournament/views.py
michael1241: I wasn't really sure what you meant on this comment but I've copied the way it works for a single season instead now with:
        career_score = 0
        career_score_total = 0
        career_perf = PerfRatingCalc()

PR #240, opened 2018-07-28
Iss227

aliquantus:
Haven't been able to test this, hopefully you can check if it looks okay.

Ensures we don't warn/set unavailable a player for non-contact, if their opponent was already unavailable.

### DIFF
--- a/heltour/tournament/automod.py
+++ b/heltour/tournament/automod.py
@@ -57,6 +57,14 @@ def withdraw_approved(instance, **kwargs):
 def automod_unresponsive(round_, **kwargs):
     groups = { 'warning': [], 'yellow': [], 'red': [] }
     for p in round_.pairings.filter(game_link='', result='', scheduled_time=None).exclude(white=None).exclude(black=None):
+        #verify that neither player is previously marked unavailable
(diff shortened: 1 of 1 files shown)


heltour/tournament/automod.py
```diff
@@ -57,6 +57,13 @@ def withdraw_approved(instance, **kwargs):
 def automod_unresponsive(round_, **kwargs):
     groups = { 'warning': [], 'yellow': [], 'red': [] }
     for p in round_.pairings.filter(game_link='', result='', scheduled_time=None).exclude(white=None).exclude(black=None):
+        #verify that neither player is previously marked unavailable
+        if round_.season.league.competitor_type == 'team':
+            white_avail, _ = PlayerAvailability.objects.get_or_create(round=round_, player=p.white)
```
cyanfish: You should avoid creating unnecessary database objects here. Instead of `get_or_create(...)`, you can use:
`filter(..., is_available=False).first()`
That will return `None` if they they are available, or an object otherwise.

Note: You can also use `get(...)` and catch the `PlayerAvailability.DoesNotExist` exception, but for whatever reason I've used the `filter` form more in heltour.

PR #265, opened 2018-11-19
Send registration confirmation email reciept

### LINKED ISSUE #250: Confirmation of registration receipt
Have a confirmation of registration receipt email in addition to confirmation of acceptance email or some other way to confirm that registration went through.

Avoids duplicate registrations and people bugging the mods about whether their registration went through.

lenguyenthanh:
Fix #250 
It's the first time I work with python, so please comment seriously about logic, code style, etc.

### DIFF
--- a/heltour/tournament/views.py
+++ b/heltour/tournament/views.py
@@ -18,6 +18,9 @@ from django.views.generic import View
 from django.utils.text import slugify
 from django.contrib.auth import login, logout
 from django.contrib.auth.models import User
+from smtplib import SMTPException
+from django.template.loader import render_to_string
+from django.core.mail import send_mail
 from ipware import get_client_ip
 
 from heltour.tournament import slackapi, alternates_manager, uptime, lichessapi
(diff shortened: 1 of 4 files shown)


heltour/tournament/views.py
```diff
@@ -621,14 +624,35 @@ def view(self, post=False):
...
         if post:
             form = RegistrationForm(self.request.POST, season=reg_season)
             if form.is_valid():
                 with reversion.create_revision():
                     reversion.set_comment('Submitted registration.')
                     form.save()
-                self.request.session['reg_email'] = form.cleaned_data['email']
+                username = form.cleaned_data['lichess_username']
+                email = form.cleaned_data['email']
+                reg = {
```
cyanfish: I believe you can replace these variables with `form.instance` and use that in place of `reg`.

heltour/tournament/templates/tournament/emails/registration_confirmation_subject.txt
```diff
@@ -0,0 +1 @@
+Registration Confirmation - {{ reg.season.league.name }}
```
cyanfish: I would prefer calling this "Registration Received" and updating the template names accordingly ("confirmation" is a bit ambiguous).

heltour/tournament/views.py
cyanfish: Looks like `email` would be used in two places. If there were more I'd say definitely use the variable; less I'd say don't; for two I have no preference.

cyanfish:
With Django errors the most relevant bit is at the bottom.
`Exception Value: Header values can't contain newlines (got u'Registration Received - Lichess4545 League\n' for header u'Subject')`
So you should be able to fix it removing the newline at the end of `registration_received_subject.txt`.

lenguyenthanh:
This issue is caused by `render_to_string('tournament/emails/registration_received_subject.txt', {'reg': reg})`

I don't know why but it adds a new line no matter what template is. It didn't happen before, but after I rebase to master upstream. Will investigate this issue when I'm back home from work.

lenguyenthanh:
Ok, the problem is caused by end of line character. Thanks @cyanfish! I pushed a fix for it.

lenguyenthanh:
oh my terrible english :(

lakinwecker:
:D

heltour/tournament/views.py
cyanfish: In case it wasn't clear what I meant here, `form.instance` is a `Registration` object that already has `season`, `lichess_username`, and `email` fields. You can use that rather than constructing your own `reg` object.

heltour/tournament/views.py
```diff
@@ -621,14 +624,37 @@ def view(self, post=False):
...
+                    'lichess_username': form.cleaned_data['lichess_username'],
+                    'email': email
+                }
+                # send confirmation of registration
+                subject = render_to_string('tournament/emails/registration_received_subject.txt', {'reg': reg})
+                msg_plain = render_to_string('tournament/emails/registration_received.txt', {'reg': reg})
+                msg_html = render_to_string('tournament/emails/registration_received.html', {'reg': reg})
+                if settings.DEBUG:
+                    print subject
+                    print reg['season'].league.name
```
cyanfish: Debug code that should be removed

heltour/tournament/views.py
lenguyenthanh: ah, yeah I misunderstood. Will fix it soon.

cyanfish:
Hey. Yeah, the code looks good now. Before I merge, can you [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) the commits together?

lenguyenthanh:
hey I did squash all commits.

PR #274, opened 2019-01-28
Trophy case

colwem:
Add a trophy case to the player profile page displaying prizes they won in previous seasons.


lakinwecker:
I'd like to see this work on lonewolf/blitz as well before we merge.  I think the css version that we talked about should be good enough to solve that.

heltour/tournament/templates/tournament/player_profile.html
```diff
@@ -193,6 +194,35 @@ <h3>Career Statistics</h3>
 			</div>
 		</div>
 		{% endif %}
+        {% if trophies %}
+		<div class="well">
+			<div class="well-head">
+				<h3>Trophy Case</h3>
```
cyanfish: If this is only showing trophies from one league, the UI should make that clear. For example, change `Trophy Case` to `{{ league.name }} Trophies`.

PR #280, opened 2019-01-31
Fix registration

colwem:
1. Users must login first
2. Users register only once
3. After first registration it says “change registration” from then on
4. Register link disappears when registration closed
5. Users who’ve had their registration rejected do not get to reregister.

The 5th might be problematic?  Easily removed.  The only register link that gets hidden right now is the one in the banner. I haven't yet tracked down all the register links/buttons.

I haven't run tests yet.  If they pass, I feel like I ought to add some tests before this is merged.


heltour/tournament/models.py
```diff
@@ -1485,6 +1501,28 @@ def other_seasons(self):
...
+        try:
+            return cls.objects.get(lichess_username=user.username, season=season).status != 'rejected'
+        except cls.DoesNotExist:
+            return True
+
+
+    @classmethod
+    def get_registration(cls, user, season):
+        try:
+            r = cls.objects.get(lichess_username=user.username, season=season)
```
lakinwecker: We should probably deal with multipleobjects as well in this case. For old seasons there will be be duplicate registrations, and if this gets called on them at any point, then we'll need to deal with it.  

I'm not sure what we should do in that case. An alternative approach is to make a migration that removes old duplicate registrations. But I'm not wild about that idea.

heltour/tournament/views.py
```diff
@@ -616,20 +616,24 @@ def view(self, username):
         pairings = player.pairings.exclude(scheduled_time=None)
         return self.ical_from_pairings_list(pairings, calendar_title, uid_component)
 
-class RegisterView(LeagueView):
+class RegisterView(LoginRequiredMixin, LeagueView):
+
```
lakinwecker: The register view should also have some protection against duplicate registrations. This will prevent things like double clicking from resulting in two registrations.  We can discuss the best way to prevent these.

heltour/tournament/models.py
colwem: I can catch it.  I tend to dislike catching things that are truly errors though.  I'm thinking thats probably not best practice in a web server though.  But we would need to report that exception in some way that blaring alarms go off for us since this should never happen.

I feel like multiple registrations really should be a gross violation and be purged/prevented.

heltour/tournament/views.py
```diff
@@ -1139,12 +1137,7 @@ def view(self):
     def _common_context(self):
         current_season_list, completed_season_list = _get_season_lists(self.league, active_only=False)
 
-        reg_season = self.season
-        if not self.season.registration_open:
-            for s in self.season.section_list():
-                if s.is_active and s.registration_open:
-                    reg_season = s
-                    break
+        reg_season = Season.get_registration_season(self.league, self.season)
```
cyanfish: This logic is wrong here; it may display a link to a completely different season. The old logic should work fine.

heltour/tournament/views.py
```diff
@@ -615,62 +614,61 @@ def view(self, username):
...
+                            html_message=msg_html,
+                        )
+                    except SMTPException:
+                        logger.exception('A confirmation email could not be sent.')
+                    self.request.session['reg_email'] = form.cleaned_data['email']
+
+                    return redirect(leagueurl('registration_success', league_tag=self.league.tag, season_tag=self.season.tag))
+            else:
+                form = RegistrationForm(instance=instance, season=reg_season)
+                if self.request.user.is_authenticated():
```
cyanfish: This check can be removed now

heltour/tournament/views.py
```diff
@@ -615,62 +614,61 @@ def view(self, username):
...
-                form.fields['classical_rating'].initial = player.rating_for(reg_season.league)
-                form.fields['has_played_20_games'].initial = not player.provisional_for(reg_season.league)
-                form.fields['already_in_slack_group'].initial = player.slack_user_id != ''
 
-        context = {
-            'form': form,
-            'registration_season': reg_season
-        }
-        return self.render('tournament/register.html', context)
+        with cache.lock(f'update_create_registration-{self.request.user.id}-{reg_season.id}'):
```
cyanfish: The registration page needs some kind of protection if the user's last registration was rejected. For now maybe just redirect to the league home page.

heltour/tournament/views.py
```diff
@@ -615,62 +614,61 @@ def view(self, username):
...
+            instance = Registration.get_registration(self.request.user, reg_season)
+            if post:
+                form = RegistrationForm(self.request.POST, instance=instance, season=reg_season)
```
cyanfish: The registration status should be reset to "pending" before saving (e.g. the user registers, is approved, withdraws, and registers again)

cyanfish: Or, perhaps better, have it create a new registration in this case.

colwem: I'm worried about what happens in the case of approved members who change their registration.  Their status would go back to pending and mods would have to re-approve them.  Is this what we want every time?  What code is triggered when the mods approve someone?

cyanfish: If an already participating player is approved (or rejected) again nothing crazy happens. It would take a deliberate action to move them off a team.
https://github.com/cyanfish/heltour/blob/master/heltour/tournament/workflows.py#L367

lakinwecker: Part of the point of this PR is to prevent duplicate registrations. Basically, once a player registers, they can't register again, they can only change their registration in some well defined ways

But getting a bunch of duplicate registrations is annoying to deal with

lakinwecker: Or am I mistaken in what this line is about

cyanfish: There needs to be some way to re-register after approval. If a user's registration is pending this will still prevent duplicates. It will also prevent re-registration in case of rejection.

To narrow down this other case (re-registration after approval), we can also block registration changes if the season player is active without any pending withdrawals.

heltour/tournament/models.py
```diff
@@ -663,6 +670,10 @@ def link_slack_account(cls, lichess_username, slack_user_id):
             signals.slack_account_linked.send(sender=cls, lichess_username=lichess_username, slack_user_id=slack_user_id)
             return True
 
+    @classmethod
+    def get_player_from_user(cls, user):
```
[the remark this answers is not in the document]
colwem: I don't believe this is actually used anymore. I can just remove it if you like.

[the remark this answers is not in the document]
cyanfish: That works too.

heltour/tournament/views.py
colwem: Soo I found lines 133-139 in forms.py.  status is already automatically set to pending every time the registration form is saved so I think this is already done.

colwem: Ok, yes it does work as expected.  Status is already reset to pending every time the form is saved.

heltour/tournament/views.py
colwem: Yeah, I honestly can't remember even doing this let alone why... Weird.

PR #283, opened 2019-02-09
Add automated create/build teams to admin

colwem:


bin/createteams.py
```diff
@@ -0,0 +1,150 @@
+# NOTE: click is not a dependency of this project, you'll have to manually
+#       install it to run this file.
+import click
```
cyanfish: Rather than commenting, wrap this in a try/catch for ImportError that `print`s this message (and then exits)

bin/createteams.py
```diff
@@ -0,0 +1,150 @@
...
+import json
+from heltour.tournament.create_teams import make_league, \
+        total_happiness, team_rating_range, team_rating_variance, reduce_variance
+
+from itertools import combinations
+from functools import partial
+
+class terminal:
```
cyanfish: Class names should be capitalized

heltour/tournament/admin.py
```diff
@@ -805,6 +790,72 @@ def player_info_view(self, request, object_id, player_name):
 
         return render(request, 'tournament/admin/edit_rosters_player_info.html', context)
 
+    def create_teams_view(self, request, object_id):
+        def get_best_league(player_data, boards, balance, count):
```
cyanfish: Can this function be moved to create_teams.py?

heltour/tournament/admin.py
```diff
@@ -30,6 +30,8 @@
 from django.contrib.staticfiles.templatetags.staticfiles import static
 from django.db.models.signals import post_save
 from django.dispatch.dispatcher import receiver
+from heltour.tournament.team_rating_utils import team_rating_range, team_rating_variance
+from heltour.tournament.create_teams import make_league, total_happiness, reduce_variance
```
cyanfish: Since the purpose of these methods isn't obvious from their names, I would prefer them to be at least partially qualified

heltour/tournament/create_teams.py
```diff
@@ -0,0 +1,386 @@
+import random
```
cyanfish: I don't love this file name; can we call it teamgen.py instead (similar to pairinggen.py)?

heltour/tournament/static/tournament/css/edit_rosters.css
```diff
@@ -36,12 +36,21 @@
 }
 
 .player {
-	display: inline-block;
+	display: flex;
```
cyanfish: The static version will also need to be incremented

heltour/tournament/templates/tournament/admin/create_teams.html
```diff
@@ -0,0 +1,41 @@
+{% extends "tournament/admin/custom_edit_workflow.html" %}
+
+{% block content %}
+<p>
+Create new teams for season {{ season }}.  This will completely delete all teams and alternates and recreate them
+according to the algorithm in `create_teams.py`.
+</p>
+
+{% if season_started %}
```
cyanfish: Please make this impossible to do once the season has started.

heltour/tournament/forms.py
```diff
@@ -344,3 +354,18 @@ def __init__(self, *args, **kwargs):
         reg = kwargs.pop('reg')
         super(MoveLateRegForm, self).__init__(*args, **kwargs)
         self.fields['prev_round'].initial = reg.round.number
+
+class CreateTeamsForm(forms.Form):
```
cyanfish: Please add a confirmation field (like [here](https://github.com/cyanfish/heltour/blob/c07d7f698794fd34c699bb43de91fb4181696cfb/heltour/tournament/forms.py#L296)) and make sure to validate it, so that it takes more than a single click from this page
e.g. "Yes, I'm sure - delete %d teams and regenerate"

heltour/tournament/admin.py
colwem: make_league specifically is a bad name. should be `generate_teams` or something like that.  I named it make league because it doesn't just create the teams it also assigns all the alternates.  But the word league has other meanings here.

heltour/tournament/templates/tournament/admin/create_teams.html
colwem: What is the purpose of the `revision` module if not to undo mistakes like this?  I figured allowing this would make testing easier.  I will now have to delete all the `rounds` in my test db to test this feature in the future.

cyanfish: Reversion works great for individual changes. It's also great for viewing a history of changes. It does not provide a good way to revert hundreds of interrelated changes which could potentially also trigger other tasks.

cyanfish: Or more generally, catastrophe prevention > catastrophe recovery.

cyanfish: If you really want you can allow it when the DEBUG flag is set like [here](https://github.com/cyanfish/heltour/blob/c07d7f698794fd34c699bb43de91fb4181696cfb/heltour/tournament/admin.py#L370).

lakinwecker: I've already made the change to prevent it when the season is started. Let's roll with this for this PR, and we can discuss how to make it easier to develop in a subsequent one

heltour/tournament/admin.py
```diff
@@ -992,8 +1043,10 @@ def team_manage_players_view(self, request, object_id):
                 purple_players.add(sp.player)
 
         expected_ratings = {sp.player: sp.expected_rating(league) for sp in season_player_objs}
+        season_started = bool(Round.objects.filter(season=season, publish_pairings=True).count())
```
cyanfish: @colwem Can you change this from `bool(...count())` to `...exists()`? I'll merge after that

lakinwecker: I thought I got both of those

lakinwecker: I'm onnit

PR #285, opened 2019-02-10
Use prev season, rather than reg form for prev alternate status.

lakinwecker:


heltour/tournament/models.py
```diff
@@ -220,15 +227,15 @@ def extract(sp):
...
-                    'previous_season_alternate': reg.previous_season_alternate,
                 })
             return info
 
         season_players = (self.seasonplayer_set
                           .filter(is_active=True)
                           .select_related('player', 'registration')
                           .nocache())
-        return [extract(sp) for sp in season_players]
+        last_season_alts = self.last_season_alternates()
```
cyanfish: You can move this definition above extract and use it as a closure variable to avoid the extra function parameter

lakinwecker: True. For whatever reason, I've been in the habit of explicit variables, rather than implicit ones these days.  Happy to do so though. 2 mins

lakinwecker: This change is pushed.

PR #298, opened 2019-03-24
Feature/bootstrap 4

pom-pom:
Hey 👋 
I ported the frontend to Bootstrap 4 & sass and added an new "Dark Mode"
Instructions how to compile:

1. `npm install -g grunt-cli` (Only needed if grunt is not installed yet)
2. `cd static-assets`
3. `npm install`
4. `grunt`
5. Optionally: Install https://chrome.google.com/webstore/detail/livereload/jnihajbhpnppcggbcgedagnkighmdlei?hl=en so the browser reloads automatically after each style change.

TODOS:
Add an option in the User Dashboard to activate/disable dark mode.
it's in `empty_base.html` Line 22: `<body class="{% if league %}theme-{{league.theme}}{% endif %} dark">`.
 The class `.dark` should be toggled by this option.

Hit me on Slack if there are any questions 👍


lakinwecker:
I'm not in favour of depending on node/grunt just for this. We should figure out a different way to do the build/dependencies for bootstrap4.

pom-pom:
@lakinwecker compilation is only done locally and the compiled CSS files are in git. 
So there would be no need to install/run anything on the server…

Here's my reasoning for using grunt:
It allows for fast compilation and live reloading in the browser when working on the styles plus some other very useful tools like auto-prefixing/minification/linting/csscomb.
grunt is also the main tool used by bootstrap themselves.
Can't really see a good alternative at the moment but very open to hear your suggestions!

lakinwecker:
I'm not debating whether grunt/npm are good tools. They are.  I'm just generally opposed to adding more dependencies if there is an easy enough way to handle them with the existing dependencies.  I'm not sure there is an easy way, but I think there probably is.  

I may not have time to look into it myself, so I'll wait until cyanfish reviews it and see if he agrees.

Also, typically static assets go into the heltour/tournament/static - is that where the build products go?

lakinwecker:
I followed your instructions and: 
```
➜  grunt
zsh: command not found: grunt
```

pom-pom:
Yes - compiled assets go there. 
We can also easily move the `static-assets` folder somewhere else to make it cleaner. 
I wasn't really sure where the best place is for the source files so I added them there.

RE grunt:
The following things are the minimal requirements for compilation:
1. sass to css compilation (obviously ¯\_(ツ)_/¯)
2. autoprefixer (needed to support older browsers and vendor prefixes - also required by bootstrap so no way around that)
3. Minification
4. CSS Comb support (although this one we could ditch, potentially if we're good/strict at writing CSS)

Optionally: Live reloading for the browser for ease of development…

pom-pom:
@lakinwecker 
If you don't have grunt you need to do
`npm install -g grunt-cli`

I'll also add to the readme and initial comment 👍

lakinwecker:
`./node_modules/.bin/grunt` also works.

pom-pom:
@lakinwecker found this: https://github.com/jrief/django-sass-processor
Seems to cover our requirements, apart from 4 and 5 which are not strictly required.

lakinwecker:
I can setup a watch command to watch and recompile scss on changes very easily. You can also use a browser plugin to reload with the .css files change as well.  

The autoprefixer thing is harder to get around. The project you linked above also requires node to be installed in order to use the autoprefixer.    Do we use that in our code, or is that only a boot strap dependency? Do we have to compile bootstrap from source, or could we pre-build it and use sass for our changes only?

lakinwecker:
If we have to install node to get autoprefixing, then I don't see a reason to avoid using other node based dependencies.  At which point the usage of grunt or not becomes a non-issue.

cyanfish:
We should be able to use a pre-built version of Bootstrap. And in general we shouldn't need prefixes if we're only supporting the browsers that lichess supports.

pom-pom:
Re: Autoprefixesr: 

We're using Bootstrap's variables to define a lot of things and compile using that there and also we can easily comment out unused things in the future. That was we don't really need to overwrite anything that isn't necessary.

Same thing goes for "our" code: It would be quite a bit of work to do all the prefixes etc manually but it would certainly be doable there.

As for using bootstrap precompiled: I'm not a big fan tbh because we would losing a lot of the flexibility and slickness what makes bootstrap a great framework.
Main plus thing of using bootstrap from source is modularity and variables and not having to override/duplicate code.

pom-pom:
> And in general we shouldn't need prefixes if we're only supporting the browsers that lichess supports.

Unfortunately that isn't the case. Even the latest versions of browsers sometimes still need prefixes for certain things. It's a bit of a jungle out there, still…

cyanfish:
Okay. My initial thoughts on this:
- Love the look. It's great.
- I can definitely do the toggle.
- There is too much going on in this PR. It will be prohibitively difficult to do a thorough review.

Ideally this would be split up into multiple PRs, for example:
1. Add the SASS tooling
2. SASSify the existing code
3. Migrate to Bootstrap 4
4. Add the dark mode

If we can get a more manageable portion of the changes to a new PR, then we can review and discuss those changes in isolation, get it merged, then repeat.

pom-pom:
@cyanfish thanks for the feedback and happy you like the look.

Heres' my feedback:
Yes - I understand it's a pretty huge PR…

Regarding the splitting: I'm not sure if there's an easy ways to split everything easily because all the parts are connected, basically – especially Point 2 and 3 (I updated to Bootstrap 4 when moving to sass so I had to do some markup changes and incorporated/adjusted and split the existing styles afterwards) 
What I could do is to make a separate PR for the tooling (Basically everything that lives in `/static-assets`, one for the Bootstrap 4  markup changes and a third for the dark theme (which is basically just one sass file…)

Would that help you manage it better?

Other option is that maybe we go through the whole thing together on a Slack call?

Let me know what the best way is for you and I'll try and do that.

pom-pom:
> I can definitely do the toggle.
I'm not sure what you mean by this

lakinwecker:
He means he will implement the user profile setting for switching to dark mode.

cyanfish:
That kind of split will definitely help. Basically I need to be able to see useful [diffs](https://github.com/cyanfish/heltour/pull/298/files).

pom-pom:
Ok. Good opportunity for me to learn how to cherry-pick! 😄 

Also I learned my lesson. I really got carried away there last night (started off just quickly wanting to add the dark theme and ended up porting the whole frontend to BS4...)
I'll probably get around to it tomorrow 👍

pom-pom:
Ok @cyanfish I split it in two:

1. #299 – Frontend Tooling / SCSS / Grunt etc.
2. #300 – Markup Changed related to Bootstrap 4

Hope this helps! I'll also add some comments to some key files/changes if I get around to it later. 👍

cyanfish:
I pulled in the dark mode css changes (as well as switching to SCSS). I left out the Bootstrap 4 + tooling changes since I don't think it provides enough value for the complexity it adds.

cyanfish:
Closing for now, let me know if you want to discuss the tooling/bootstrap changes more

PR #299, opened 2019-03-25
Feature/frontend tooling

pom-pom:
As discussed, here's the first chunk.
It includes all the Styles / Sass Compilation / Grunt


colwem:
This appears to be almost entirely commiting a library to this project.  If this were python you would never do this. You would add it as a dependency in the requirements.txt and then pip install it.  Is there no similar system for adding bootstrap as a dependency?

pom-pom:
@colwem we could use Yarn or NPM to pull bootstrap as a dependency, yeah but would need some refactoring on my part because we use variables and other things.

I would suggest that we start like that now and I'll implement this later (we had bootstrap as a css and JS file in the repo before so I don't see a big difference in regards to code size now but I get your point and think I'll be able to come up with a solution at a later stage 👍)

lakinwecker:
You can also use git submodules or, preferably, use the fabfile to get the dependency directly from git.

pom-pom:
Hey @lakinwecker @cyanfish 👋 

So do I need to implement Bootstrap as a dependency in order for my PR to get merged?
As it was before my changes, Bootstrap was also checked in the repo directly, so I think adding it as a dependency would be an improvement/change that I could do at a later point. But if there's no way around it, I can try and do it now I guess.

Or is it more the case that you just didn't get a chance to check the PR yet? (which is 100% fine! I know you have lots of other, more important things to do)

Anyways would be nice to get a quick status update 👍 

Thanks 1000

lakinwecker:
I haven't had a chance to work on it yet. Maybe this weekend?

pom-pom:
Ok great 👍 thanks for the feedback. No rush - just wanted to check quickly because I wasn't sure how to proceed.

lakinwecker:
Sorry. I didn't get to this over the weekend. Worked all day yesterday instead.

pom-pom:
Been there … Sunday work is the worst…

No worries. it's not like this feature is the most pressing – it's just hard for me to look back, after being touched by darkness 😃

cyanfish:
Closing for now, let me know if you want to discuss the tooling/bootstrap changes more

PR #301, opened 2019-03-29
Upd notify - add edit availability url

Kraaft:
-NOT SURE ABOUT MYSELF-

It is supposed to add a message that explain how to set yourself available (for the common case where someone has been late and want to stop the alternate search and play this round.)

I had to create the url via abs_url(reverse()) and if I got it right: edit_availability needs no argument.
I hope the syntax is correct though (meaning the 3 elements in <reverse> will not require 3 elements in <args=> cause I put only two.)

Please check that I've been a good boy. Thanks.

### DIFF
--- a/heltour/tournament/notify.py
+++ b/heltour/tournament/notify.py
@@ -638,9 +638,11 @@ def notify_unresponsive(round_, player, punishment, allow_continue, pairing, **k
     season = round_.season
     league = season.league
     appeal_url = abs_url(reverse('by_league:by_season:modrequest', args=[league.tag, season.tag, 'appeal_late_response']))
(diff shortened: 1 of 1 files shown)


lakinwecker:
Based on a cursory review, I don't see any reason why this won't work.

lakinwecker:
I'm not sure how to test it easily, so I don't have time to test it today, but I'll look into it as soon as I can.

cyanfish:
This is only applicable for the team league, but the method is also used for lonewolf. You'll want to put this behind a condition (like allow_continue below).

cyanfish:
Thanks!

PR #306, opened 2019-04-29
Fix registration bug due to username mismatches

colwem:


colwem:
Considering registration is open, and this is a bug fix for something right now, can we move on this ASAP?

PR #312, opened 2019-05-12
Add user to Registration and Player model

colwem:
This gets rid of all the "Joining" logic between `Player` `User` and `Registration` that is based on usernames.  Instead from now on `Player` and `Registration` have a `user` `ForeignKey`.  The `user` field should be not nullable but I don't know how to ensure that all rows in the User and Player table can be paired.

I'm sure I missed a few things but missing a few things shouldn't make things break.  It just means some stuff is still joining on username.  Tests pass.


heltour/tournament/models.py
```diff
@@ -1553,7 +1564,7 @@ def other_seasons(self):
         return SeasonPlayer.objects.filter(player__lichess_username__iexact=self.lichess_username).exclude(season=self.season)
 
     def player(self):
-        return Player.objects.filter(lichess_username__iexact=self.lichess_username).first()
+        return self.user.player
```
colwem: woops, this is wrong.  Should use `Player.get_or_create(self.user)` or something

heltour/tournament/models.py
```diff
@@ -693,7 +695,11 @@ def update_profile(self, user_meta):
 
     @classmethod
     def get_or_create(cls, lichess_username):
-        player, _ = Player.objects.get_or_create(lichess_username__iexact=lichess_username, defaults={'lichess_username': lichess_username})
+        try:
+            user = User.objects.get(username__iexact=lichess_username)
+        except User.DoesNotExist:
+            raise f'no user with lichess username {lichess_username} exists'
```
lakinwecker: I don't expect a `get_or_create` method to raise an error when something doesn't exist. I expect it to create that object.  I haven't analyzed where this method is used well enough to know if that makes sense everywhere, but that's what I would expect from a method named like this.

colwem: It's a constraint violation.  Don't we expect a constraint violation in a method called create to throw an exception

lakinwecker: I don't see it as a constraint violation. I see it as a request to ensure that a Player with that username exists in the system. If the player doesn't exist, we create it. If the Player requires that the user also exists, then you should create both.

heltour/tournament/views.py
```diff
@@ -1833,13 +1833,14 @@ def view(self, secret_token=None, post=False):
                             user = User.objects.create_user(username=token.lichess_username, password=create_api_token())
                     user.backend = 'django.contrib.auth.backends.ModelBackend'
                     login(self.request, user)
+                    player = Player.get_or_create(user)
```
colwem: This should be where all player objects get created now.

cyanfish:
Referencing the User model in more places isn't the approach I want to take:
- The only unique information in the User model is related to auth, which is only relevant for a few isolated portions of the code. Most of the time you want data from the Player model, so it's better to standardize on that. In views you can reference self.player as a convenience (which isn't always done, so there's potential for cleanup).
- If we go to the full extent of denormalizing the username field, that has significant performance risks. Since usernames are used _everywhere_, that means lots of fetches. Cacheops helps but is still much slower than just having the field directly on the player model.

Feel free to revise this PR (or create another) to add a Player field to Registration. That plus cleaning up some views that can use self.player should get rid of all unnecessary username comparisons except in a few bits of auth-related code.

lakinwecker:
This branch is older and out of date. It was a good idea, and I'd still like to see it done, but as it stands it would need cleanup. Closing for now, feel free to open a new one.

PR #314, opened 2019-05-18
Removes the lichess_username and some other changes

lakinwecker:
This is based off of colwem's PR for adding a OneToOne field to Registration/Player - but includes a few other things:

1. It updates the migration to properly create Users for all Registration/Player records that have a non-empty username.  This might clutter the User records a little, but I think that's fine.
2. It removes the database field for lichess_username on the Player record as the user OneToOneField should be enough. It leaves the lichess_username on the Registration record as it seems appropriate to record it there.
3. I replaces lichess_username with a property on Player as a shortcut to getting the username in templates etc.
4. It updates all of the queries to refer to `user__username` as appropriate.
5. It replaces the "Change Registration" link with an "Edit Availability" link when the season is started and the registration is already approved. I don't see why someone would need to edit their registration in this case.

This might make the PR harder to review, as it's far more changes. I'm happy to bring it up to date if you want to merge colwem's branch first and then deal with this as as second pass.


cyanfish:
See https://github.com/cyanfish/heltour/pull/312#issuecomment-568286210

PR #352, opened 2020-08-26
Improve lichess api response handling

lakinwecker:
I have tested it yet, but I'm going to try to figure it out.

### DIFF
--- a/heltour/api_worker/views.py
+++ b/heltour/api_worker/views.py
@@ -35,6 +35,16 @@ def _do_lichess_api_call(redis_key, path, method, post_data, params, priority, m
         else:
             r = requests.get(url, params, headers=headers)
 
+        if r.status_code == 429:
+            logger.info('API 429, sleeping for a minute')
+            time.sleep(60) # Abide by the API rules
+
+        if r.status >= 400 and r.status <= 500:
+            logger.info('API Client Error: %s: %s', r.status, r.text)
+            cache.set(redis_key, f'CLIENT-ERROR: {r.text}', timeout=60)
+            time.sleep(2)
+            return
+
         if r.status_code == 200:
             # Success
             logger.info('API success')
--- a/heltour/api_worker/views.py
+++ b/heltour/api_worker/views.py
@@ -35,10 +35,6 @@ def _do_lichess_api_call(redis_key, path, method, post_data, params, priority, m
         else:
             r = requests.get(url, params, headers=headers)
 
-        if r.status_code == 429:
-            logger.info('API 429, sleeping for a minute')
-            time.sleep(60) # Abide by the API rules
-
         if r.status_code >= 400 and r.status_code <= 500:
             logger.info('API Client Error[url:%s]: %s: %s', url, r.status_code, r.text)
             cache.set(redis_key, f'CLIENT-ERROR: {r.text}', timeout=60)
@@ -69,6 +65,7 @@ def _do_lichess_api_call(redis_key, path, method, post_data, params, priority, m
                           params, priority, max_retries, format, retry_count + 1)
 
     if r is not None and r.status_code == 429:
+        logger.warning('API 429, sleeping for a minute')
         # Too many requests
         time.sleep(60)
     else:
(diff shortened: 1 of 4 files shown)


lakinwecker:
This PR adds handling for lichess 429 response status codes, and also otherwise distinguishes between client side errors.

It also uses a specific client side error to determine when a player has blocked us (specifically or inadvertently). And if they have, it turns of lichess notifications for them.

Finally, when sending lichess emails, we ignore all client side errors.

heltour/api_worker/views.py
```diff
@@ -35,6 +35,16 @@ def _do_lichess_api_call(redis_key, path, method, post_data, params, priority, m
         else:
             r = requests.get(url, params, headers=headers)
 
+        if r.status_code == 429:
```
cyanfish, line 38: This is supposed to be handled below (line 61 of the original file)

heltour/tournament/lichessapi.py
```diff
@@ -36,13 +36,18 @@ def _apicall(url, timeout=300, check_interval=0.1, post_data=None):
         if time_spent >= timeout:
             raise ApiWorkerError('Timeout for %s' % url)
 
+def _apicall_with_error_parsing(*args, **kwargs):
+    result = _apicall(*args, **kwargs)
+    if result == '':
+        raise ApiWorkerError('API failure')
+    if result.startswith("CLIENT-ERROR: "):
+        raise ApiClientError(f'API failure: {result}')
```
cyanfish, line 44: There's one spot where ApiWorkerError is caught (tasks.py), can we fix/verify that to make sure it still catches the right exception?

heltour/tournament/notify.py
```diff
@@ -36,7 +36,17 @@ def _message_multiple_users(league, usernames, text):
...
-        lichessapi.send_mail(username, subject, text)
+        try:
+            lichessapi.send_mail(username, subject, text)
+        except lichessapi.ApiClientError as e:
+            reason = str(e)
+            if "The message was rejected" in reason:
+                # If they have blocked us specifically, or if they have
+                # Turned off messages from non-followers In that case,
+                # make an attempt to turn off their lichess noifications.
+                PlayerNotificationSetting.objects.filter(
```
cyanfish, line 47: Do we need to do this (vs. just failing gracefully)? It's another step for the user to enable lichess messages if they choose to do so.

heltour/api_worker/views.py
lakinwecker: :+1:

heltour/tournament/lichessapi.py
lakinwecker: Will do

heltour/tournament/notify.py
lakinwecker: Seems silly to keep trying to send lichess messages to users that are blocking it.  It's not strictly necessary, but it's a wasted request and eats into our request budget for no real gain.  I'm in favour of it, which is why I added it, but if you don't like it, I can remove it.

heltour/api_worker/views.py
lakinwecker: Done.

heltour/tournament/lichessapi.py
lakinwecker: I made ApiClientError inherit from the ApiWorkerError, which should allow any usages of except ApiWorkerError to continue to work.

heltour/tournament/notify.py
cyanfish: This won't actually work in most cases, because unless the player has made a manual change their PlayerNotificationSetting is [implicit](https://github.com/cyanfish/heltour/blob/3b1ac21e616fbbdbe27164f06cb57b94f7f98dba/heltour/tournament/models.py#L2524). You can try and create a set of objects for the current league, but I'd recommend just removing it since it's a small % of users.

lakinwecker: Ah, ok.   I'll remove it.

lakinwecker: Removed.

PR #354, opened 2020-09-07
Fiddling with invite users

lakinwecker:
I'm pretty sure both of these were live at some point, because I made them in master and forgot to remove them the last time I deployed. I think both are worthy of being live.


lakinwecker:
Also, I have time to deal with the fallout as well.

lakinwecker:
Sorry, I thought this also had the reduction by 4 of api calls, but it seems that one made it into your master as well. Deploying this one is not as important.

PR #360, opened 2020-10-18
Adding assigned alts to lookup for team by player

lakinwecker:
I haven't tested yet, but should fix it, right?


lakinwecker:
YOLO

PR #375, opened 2021-04-11
WIP - initial gdpr erase command.

lakinwecker:
Not ready for merge yet, but I have some questions, so I thought I'd send it.


heltour/tournament/views.py
```diff
@@ -688,7 +688,8 @@ def lone_view(self, round_number=None, team_number=None):
 
 class ICalPlayerView(BaseView, ICalMixin):
     def view(self, username):
-        player = get_object_or_404(Player, lichess_username__iexact=username)
+        player = get_object_or_404(
+            Player, lichess_username__iexact=username, gdpr_erased=False)
```
ioparaskev, line 692: since the players information will be scrubbed, do we need this filter? Why not get the player and have a hint that the information is gdpr scrubbed? (maybe it is a GDPR restriction that's why I'm asking :P)

lakinwecker: Because this page will list out all of the games they played and teams they were part of.  This allows someone who is interested to work out who they were.  The scrubbing is intended to stop that. So we made the decision to remove this page to make it harder to work out who they are.

PR #380, opened 2021-04-18
A command to cleanup the comments.

lakinwecker:


heltour/tournament/management/commands/cleanupcomments.py
```diff
@@ -0,0 +1,49 @@
...
+            "playerwithdrawal",
+            "playerwarning",
+        ]
+        ct_pks = [
+            ct.pk for ct in ContentType.objects.filter(model__in=models)
+        ]
+        assert(len(ct_pks) == len(models))
+        for badword in ['mark', 'cheat', 'alt', 'tos violation']:
+            Comment.objects.filter(
+                content_type_id__in=ct_pks,
```
cyanfish, line 42: Is there a reason to restrict to these particular content types?

lakinwecker: The idea was to prevent comments which are unlikely to have personal information in them from being deleted. So like, comments on a season, or a document

lakinwecker:
I've updated this based on a new discussion. The new discussion says don't try and heuristically determine which comments to delete. Instead, delete everything from pre-2021 for these content types which were not made by the "System" user.

PR #396, opened 2021-06-26
Update to new PKCE OAuth Flow 

niklasf:
for ornicar/lila#9214


niklasf:
https://github.com/Lichess4545/4545vagrant/issues/6 masked
```
django.db.utils.IntegrityError: null value in column "refresh_token" violates not-null constraint
```
during testing. Will also have to remove that column (which is unused anyway).

PR #398, opened 2021-08-10
Fix the space before player's rating in team profile page

timothyha:
Example of the wrong formatting, https://www.lichess4545.com/team4545/season/27/team/35/

Board 1: | My_goal_2400 ( 2202)

There's a space before 2202.


lakinwecker:
Looks good to me.

PR #404, opened 2021-10-07
fixes #402 - closed accounts and accounts with tosViolation are detected again

### LINKED ISSUE #402: Account status not updated anymore
If a player closes their account or gets marked, their accout status is not updated anymore. This has to do with changes in the lichess user API: 
- Previously, closed accounts had the "disabled": true tag. The tag is now called "closed".
Example:  `{"id":"redacted","username":"redacted","closed":true}`
- Previously, engine users had the "engine: True" tag and boosters had the "booster: True" tag. Now both these flags have the "tosViolation": true tag.

Initial investigation indicates that the following lines (and the underlying player model) have to be updated to fix the issue.
https://github.com/cyanfish/heltour/blob/748a011c71ca07375f7892007d6f1356e2abbbf5/heltour/tournament/models.py#L794-L796

hopffgam:
During the investigation I found two root causes that are fixed in this changeset
1. lichess api does not provide the engine and booster tags anymore,we only get tosViolation. I have adapted the mappings and the allowed values for Player.account_status.
2. lichess api does not provide a 'perfs' tag for closed accounts, which leads to an exception in the Player.update_profile when we try to get the classical information from 'perfs'. This changeset handles the exception now.


heltour/tournament/migrations/0192_player_account_status.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 2.2.13 on 2021-10-08 05:40
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    
+    def update_player_status(apps, schema_editor):
+        Players = apps.get_model('tournament', 'Player')
+        for player in Players.objects.all():
```
lakinwecker, line 10: This is better written:
`Players.objects.filter(account_status__in=('engine', 'booster')).update(account_status='tos_violation')`

as this will be a single query.

lakinwecker:
Aside from the minor comment about issuing a single query to update the records, this seems fine for me. @cyanfish ?

heltour/tournament/migrations/0192_player_account_status.py
hopffgam: done

heltour/tournament/models.py
```diff
@@ -787,14 +786,18 @@ def save(self, *args, **kwargs):
 
     def update_profile(self, user_meta):
         self.profile = user_meta
-        classical = user_meta['perfs'].get('classical')
```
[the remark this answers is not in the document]
hopffgam: done

lakinwecker:
Awesome, looks good to me.

PR #425, opened 2022-01-24
More aggressive wrapping for team names in tables

zbidwell:
Long unbreakable words in team names (e.g. a FEN string...) would either overflow or cause tables to grow too wide for narrow screens.

This CSS change just allows for breaking of normally unbreakable words in .team-links in tables.

I can't run vagrant here, and I don't have anything that can run safari, so this was just tested with the inspector in chrome and firefox. I can't imagine how this could break anything, but it might be worth the time to double check anyways before pushing to prod.

Example of the issue on an iOS device, courtesy of M0r1:
long team name example

### DIFF
--- a/heltour/tournament/static/tournament/css/_common.scss
+++ b/heltour/tournament/static/tournament/css/_common.scss
@@ -130,6 +130,13 @@ div.inline {
     color: inherit;
 }
 
+table .team-link {
+    // Both lines have the same effect. `anywhere` value is not supported on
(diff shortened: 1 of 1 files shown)


lakinwecker:
I'm also not testing this, and I'm going to trust you on the CSS, because CSS is not my strong suit.  Finally, this has very little possibility of having major impact, so it's safe to YOLO it up. Let's go. :shipit:

Issue #335, opened 2020-03-11
Red card updates


michael1241:
Set player unavailable for the rest of the season and lock editing so player cannot set themselves available again.
Related: withdrawn players should also be set unavailable for the rest of the season automatically.

Unclevinny:
It'd be nice to notify the team captain as well.

mathgeek:
Can close this I think

PR #426, opened 2022-01-30
335 red card handling

hopffgam:
Automated handling of red cards for team league
- set player unavailable for current and remaining rounds
- Players cannot change their availability when they have a red card
- Captains connot change availability of players with red card
- However, moderators can still change the availability via the tournament admin.

### DIFF
--- a/heltour/tournament/models.py
+++ b/heltour/tournament/models.py
@@ -1794,6 +1794,13 @@ class SeasonPlayer(_BaseModel):
         super(SeasonPlayer, self).__init__(*args, **kwargs)
         self.initial_unresponsive = self.unresponsive
         self.initial_player_id = self.player_id
+        self.initial_games_missed = self.games_missed
+        
+    def _set_unavailable_for_season(self):
+        rounds = Round.objects.filter(season=self.season, is_completed=False)
+        for r in rounds:
+            PlayerAvailability.objects.update_or_create(round=r, player=self.player,
+                                                    defaults={'is_available': False})
 
     def player_rating_display(self, league=None):
         if self.final_rating is not None:
@@ -1816,6 +1823,9 @@ class SeasonPlayer(_BaseModel):
             if alt.priority_date_override is None or alt.priority_date_override < current_date:
                 alt.priority_date_override = current_date
                 alt.save()
+                
+        if self.games_missed >= 2 and self.initial_games_missed < 2:
+            self._set_unavailable_for_season()
 
         super(SeasonPlayer, self).save(*args, **kwargs)
 
--- a/heltour/tournament/views.py
+++ b/heltour/tournament/views.py
@@ -1828,6 +1828,7 @@ class AvailabilityView(SeasonView, LoginRequiredMixin):
 
             availability_set = PlayerAvailability.objects.filter(player__in=player_list,
                                                                  round__in=round_list).nocache()
+            season_player_set = SeasonPlayer.objects.filter(player__in=player_list, season=self.season).nocache()
             is_available_dict = {(av.round_id, av.player_id): av.is_available for av in
                                  availability_set}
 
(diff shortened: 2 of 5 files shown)


heltour/tournament/models.py
```diff
@@ -1816,6 +1823,9 @@ def save(self, *args, **kwargs):
             if alt.priority_date_override is None or alt.priority_date_override < current_date:
                 alt.priority_date_override = current_date
                 alt.save()
+                
+        if self.games_missed >= 2 and self.initial_games_missed < 2:
+            self._set_unavailable_for_season()
```
lakinwecker, line 1828: I'm not certain how the `initial_games_missed` interacts with the `games_missed`? Does `games_missed` always include all of the `initial_games_missed`?  If so, does this mean that missing 2 games (1 at the beginning and 1 elsewhere) you'll get set to unavailable for the entire season?

heltour/tournament/views.py
```diff
@@ -1839,14 +1840,16 @@ def view(self, post=False):
                     for p in player_list:
                         field_name = 'av_r%d_%s' % (r.number, p.lichess_username)
                         is_available = self.request.POST.get(field_name) != 'on'
-                        if is_available != is_available_dict.get((r.id, p.id), True):
+                        
+                        if (is_available != is_available_dict.get((r.id, p.id), True) and 
+                        SeasonPlayer.objects.filter(player=p, season=self.season).first().card_color != "red"):
```
lakinwecker, line 1845: `first()`  will sometimes return `None`. Are we sure that this will never error?

heltour/tournament/views.py
```diff
@@ -1839,14 +1840,16 @@ def view(self, post=False):
...
                             PlayerAvailability.objects.update_or_create(player=p, round=r,
                                                                         defaults={
                                                                             'is_available': is_available})
                 return redirect('by_league:by_season:edit_availability', self.league.tag,
                                 self.season.tag)
 
-            round_data = [(r, [(p, is_available_dict.get((r.id, p.id), True)) for p in player_list])
+            round_data = [(r, [(p, is_available_dict.get((r.id, p.id), True), season_player_set.get(player=p).card_color=="red") for p in player_list])
```
lakinwecker, line 1852: get will throw an exception if this player doesn't exist, are we sure this will never error?

lakinwecker: Additionally here, this will issue a single query for every player.  Perhaps we should actually query for all of the red cards, setup a dict `{<player_id>: <card_color>}` and then look up in here? It would both increase safety and performance.

lakinwecker (changes requested):
In general, this looks good. Logic is sound, and I think it will implement the desired behavior. I just worry that we've introduced a few more place to cause 500 errors and I want to be certain that we address those.

heltour/tournament/models.py
hopffgam: `games_missed` is a property of a season player. It is used to determine if a season player has a yellow or red card. If a season player has `games_missed = 1` it means that they have a yellow card in the respective season. If they have `season_games >= 2` they have received a red card. The idea is that once a player receives a red card, they are set unavailable for the remainder of the season. So far the team league mods had to do that manually: Each time a player received a red card (missing games went from 1 to 2) we had to notice that and had to set the player unavailable for the remaining rounds. 

Now, the only way that `games_missed` could be lower than `initial_games_missed` is when a moderator or a succesful no-show appeal reduce the number of games missed. In that case we would still have to ask the player to update their availability manually, as we are not able to determine which rounds we should set back to avialable (i.e. which rounds heltour set to unavailable because of the red card and which rounds the player had set to unavailable already before).

Not sure if that answers your question. TLDR: `games_missed` always includes `initial_games_missed`, unless the player gets a card revoked by moderator action. Either directly on the season player object or by accepting a no-show appeal.

heltour/tournament/views.py
hopffgam: It _should_ not error from the context when the view is called. But I will submit a commit that handles it.

heltour/tournament/views.py
hopffgam: Most recent commit uses dictionary and accesses the season player objects in safer way

hopffgam:
last commit is my attempt to address the requested changes

heltour/tournament/views.py
lakinwecker: This is much better. Thanks!

heltour/tournament/models.py
lakinwecker: The `season_games >= 2` should be `games_missed >= 2` right?

lakinwecker: Oh, I missed that `initial_games_missed` is a cached version of the value from when the object is created. 

We could probably make this slightly clearer by refactoring that into a method of some sort:
`def just_received_red_card(self):` and have it check them, with a minor comment to that effect. 

But it's not important, I'm merging.

mathgeek:
Fixes #335

PR #438, opened 2022-02-13
Cleanup

mathgeek:
image

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -46,8 +46,6 @@ def get_password():
 
 # -------------------------------------------------------------------------------
 # Figure out where we are on the directory system
-import os.path
-import sys
 
 relative_path_to_this_module = os.path.dirname(os.path.abspath(sys.modules[__name__].__file__))
 absolute_path_to_this_module = os.path.abspath(relative_path_to_this_module)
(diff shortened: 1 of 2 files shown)


fabfile.py
```diff
@@ -89,7 +87,6 @@ def update():
         python_dependency(repo.name, PYTHON_VERSION)
 
     python_dependency('heltour', PYTHON_VERSION)
-    local("pip install -r {}".format(project_relative("requirements.txt")))
```
mathgeek, line 92: I'm seeing two of these failures (see screenshot) - not sure if it's just this one or if one of the calls above it is also trying to install the old requirements.txt. Guess I could try standing up the vagrant again and see what it does

fabfile.py
```diff
@@ -89,7 +87,6 @@ def update():
         python_dependency(repo.name, PYTHON_VERSION)
 
     python_dependency('heltour', PYTHON_VERSION)
```
mathgeek, line 89: @lakinwecker Is it necessary to have this and the above "container" at the same time? Also not entirely sure what purpose these serve

fabfile.py
mathgeek: image

Looks good

fabfile.py
lakinwecker: @mathgeek These jam a `.pth` file into the virtual environment which allows python to find the modules appropriately.

lakinwecker: And yes, they're required.

lakinwecker (approved):

PR #451, opened 2022-08-07
fixes #435

### LINKED ISSUE #435: Alternate searches should not start automatically if a new opponent was assigned after round start
In the team league after 25 hours each pairing is checked for unresponsive players and alternate searches are started for players that are determined to be unresponsive. The problem is that if a player was waiting for an alternate to be assigned as an opponent, and that alternate gets assigned close before the check, the player only has very little time to contact their opponent to avoid being marked as unresponsive.

This should not happen. As a short term fix, players should probably not be considered unresponsive by our software at all if their opponent has been assigned after round start. Long term fix could be to set a player as unresponsive if they did not contact their opponent within 24 hours after the opponent was assigned to the pairing.

hopffgam:
timestamp when player is changed on a pairing is recorded.
when player was added to pairing after round start we do not trigger an
alternate search in the team league


hopffgam:
closing this as I screwed up too much

PR #452, opened 2022-08-08
435 do not start alt search for late opponent assignment

### LINKED ISSUE #435: Alternate searches should not start automatically if a new opponent was assigned after round start

hopffgam:
fixes #435
introduce a timestamp for when a player was changed last time on a pairing.
Do not trigger unresponsive workflows if a player change was registered after the round start and the pairing is for the team league, as that means that an alternate was put on the pairing late and the opponent in fact did not have 24 hrs to respond.


heltour/tournament/models.py
```diff
@@ -16,6 +16,7 @@
 from django_comments.models import Comment
 from heltour import settings
 import reversion
+from _datetime import date
```
lakinwecker, line 19: Typically it's not a good idea to depend directly on the underlying C libs, which usually start with an underscore. And it's best to depend on the main python one, which does not start with an underscore. 

Is there a reason why you are using the underscore version?

heltour/tournament/automod.py
```diff
@@ -76,12 +76,13 @@ def automod_unresponsive(round_, **kwargs):
...
         white=None).exclude(black=None):
         # verify that neither player is previously marked unavailable
+        # verify that both players on the pairing have been assigned more than 24 hours ago.
         if round_.season.league.competitor_type == 'team':
             white_unavail = PlayerAvailability.objects.filter(round=round_, player=p.white,
                                                               is_available=False).exists()
             black_unavail = PlayerAvailability.objects.filter(round=round_, player=p.black,
                                                               is_available=False).exists()
-            if white_unavail or black_unavail:
+            if white_unavail or black_unavail or p.paring_changed_after_round_start():
```
lakinwecker, line 85: `paring` -> `pairing`

heltour/tournament/models.py
```diff
@@ -1519,6 +1523,12 @@ def get_player_presence(self, player):
             presence = PlayerPresence.objects.create(pairing=self, player=player,
                                                      round=self.get_round())
         return presence
+    
+    def paring_changed_after_round_start(self):
```
lakinwecker, line 1527: `paring` -> `pairing`

heltour/tournament/models.py
hopffgam: No, that was either an auto-import or I saw this in another heltour file. I will change that.

lakinwecker:
I'm going to just describe my understanding of this change, to make sure that I understand it:

1. The previous check (just `white_unavail or black_unavail`) worked when the 24 hours contact period was over and an alternate had not yet been assigned.
2. This change is to address the situation where an alternate is assigned at hour 23 and the player that was waiting for a pairing only has an hour to respond?

heltour/tournament/models.py
```diff
@@ -1423,6 +1424,9 @@ class PlayerPairing(_BaseModel):
     game_link = models.URLField(max_length=1024, blank=True, validators=[game_link_validator])
     scheduled_time = models.DateTimeField(blank=True, null=True)
     colors_reversed = models.BooleanField(default=False)
+    
+    #We do not want to mark players as unresponsive if their opponents got assigned after round start
+    last_time_player_changed = models.DateTimeField(blank=True, null=True)
```
lakinwecker, line 1429: To be consistent, we should probably use a name that starts with `date_`, like the `date_modified` or `date_created` attributes.  Maybe something like:
`date_players_modified`

lakinwecker:
Just minor things with this one. But otherwise looks like it should work as intended.

lakinwecker:
Looks good. I'll merge and deploy both of these in the upcoming days

lakinwecker (commented):
LGTM

PR #453, opened 2022-08-10
fixes #450

### LINKED ISSUE #450: Players should not be able to set themselves unavailable when their game is already scheduled
Affects only the team league as in the individual competitions players cannot change the availability of the ongoing round anyway. Related to #236 but independent. #236 handles the situation where a player is set unavailable and then the game is scheduled, this issue handles the situation when the game is scheduled and then the player tries to go unavailable.

hopffgam:
fixes #450 
if players already have a scheduled game for a round they cannot set
themselves unavailable anymore.
This only affects the team league directly, as in lonewolf and chess960
you cannot set yourself unavailable for an ongoing round anyway.

### DIFF
--- a/heltour/tournament/api.py
+++ b/heltour/tournament/api.py
@@ -375,7 +375,6 @@ def set_availability(request):
         round_ = _get_next_round(league_tag, season_tag, round_num)
         player = Player.objects.filter(lichess_username__iexact=player_name).first()
         season_player = SeasonPlayer.objects.filter(player=player, season=round_.season).first()
-        
     except IndexError:
         return JsonResponse({'updated': 0, 'error': 'no_matching_rounds'})
 
--- a/heltour/tournament/api.py
+++ b/heltour/tournament/api.py
@@ -375,6 +375,7 @@ def set_availability(request):
         round_ = _get_next_round(league_tag, season_tag, round_num)
         player = Player.objects.filter(lichess_username__iexact=player_name).first()
         season_player = SeasonPlayer.objects.filter(player=player, season=round_.season).first()
+        
     except IndexError:
         return JsonResponse({'updated': 0, 'error': 'no_matching_rounds'})
 
@@ -386,6 +387,12 @@ def set_availability(request):
     
     if season_player.card_color == "red":
         return JsonResponse({'updated': 0, 'error': 'player_red_card'})
+    
+    if season_player.has_scheduled_game_in_round(round_):
+        return JsonResponse({'updated': 0, 'error': 'scheduled_game'})
+    
+    if has_scheduled_game:
+        return JsonResponse({'updated': 0, 'error': 'game_scheduled'})
 
     PlayerAvailability.objects.update_or_create(round=round_, player=player,
                                                 defaults={'is_available': is_available})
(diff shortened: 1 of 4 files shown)


heltour/tournament/api.py
```diff
@@ -375,6 +375,7 @@ def set_availability(request):
         round_ = _get_next_round(league_tag, season_tag, round_num)
         player = Player.objects.filter(lichess_username__iexact=player_name).first()
         season_player = SeasonPlayer.objects.filter(player=player, season=round_.season).first()
+
```
lakinwecker, line 378: Why the extra line?

hopffgam: fixed

heltour/tournament/api.py
```diff
@@ -386,6 +386,12 @@ def set_availability(request):
     
     if season_player.card_color == "red":
         return JsonResponse({'updated': 0, 'error': 'player_red_card'})
+    
+    if season_player.has_scheduled_game_in_round(round_):
+        return JsonResponse({'updated': 0, 'error': 'scheduled_game'})
+    
+    if has_scheduled_game:
```
lakinwecker, line 393: Where is this variable set? Why is the error for this clause different from the error in the one above?

heltour/tournament/templates/tournament/availability.html
```diff
@@ -45,13 +45,13 @@ <h3>Edit Availability</h3>
                                 <tr>
                                     <td>Round {{ r.number }}</td>
                                     <td>{{ r.start_date|date_or_q:'%b %-d' }} - {{ r.end_date|date_or_q:'%b %-d' }}</td>
-                                    {% for p, is_av, is_red in av_list %}
+                                    {% for p, is_av, is_fixed in av_list %}
```
lakinwecker, line 48: I don't remember what this variable was supposed to indicate in the previous version, but the new name also doesn't seem particularly clear to me. 

Is it necessary for this branch?

heltour/tournament/views.py
```diff
@@ -1839,14 +1839,20 @@ def view(self, post=False):
             season_player_set = SeasonPlayer.objects.filter(player__in=player_list, season=self.season).nocache()
             has_red_card_dict = {sp.player : sp.card_color == "red" for sp in season_player_set}
             
+            if self.league.competitor_type == 'team':
```
lakinwecker, line 1842: We use this kind of if statement twice in two different places, probably a candidate for refactoring. 

```python
def is_team_league(self):
    ...
``` or something.

lakinwecker:
Pointed out one thing that I believe is a bug, and a few other minor cleanups that I think are worth doing. Ping me when they're done and I'll review it again.

heltour/tournament/templates/tournament/availability.html
hopffgam: That parameter controls if the availability checkbox can be changed or not. I have renamed it to "cannot_change" in the upcoming commit. I know that using negations in boolean variables is usually not very elegant but here I think it conveys the usage.

heltour/tournament/views.py
hopffgam: I defined that function and and replaced those competitor_type=='team' calls, but I limited myself to those files that I had to touch anyway (views.py and models.py)

heltour/tournament/templates/tournament/availability.html
lakinwecker: What about `is_readonly` or `is_immutable`.  I don't mind `cannot_change`, but these are other potential names. I'm happy to merge it as is, though.

hopffgam: I like is_immutable bebest and will adapt it.

PR #455, opened 2022-09-02
contact is for people not on slack

### LINKED ISSUE #454: Contact Form is only for people not on slack.
Make it clear on the contact form that people already on slack should ask their questions there.

glbert-does:
closes  #454


glbert-does:
wording is of course debatable. but imo we get too many questions via e-mail from people already on slack. usually questions that could easily answered by anyone in #general.

lakinwecker (approved):

PR #458, opened 2022-09-15
fixes #236

### LINKED ISSUE #236: Set player available if a game is scheduled
While an alt search is in progress, the player is still on the pairing and can schedule a game. This does not cancel the alt search. If an alt is subsequently found, the previous game scheduling will be removed (and the alt can schedule). In particular, this has happened when a player is automatically set unavailable after 24h but quickly schedules the game after that, and forgets to update their availability status.

Two solutions:

a) If a game is scheduled, set the player available for that week. This cancels the alt search. 
b) Only allow available players to schedule. (Trigger chesster message when trying to schedule.)

hopffgam:
fixes #236 if a game gets scheduled, we set the related availability objects to is_available=True if they exist.


heltour/tournament/models.py
```diff
@@ -1583,7 +1584,19 @@ def save(self, *args, **kwargs):
                 old_black_setting = PlayerNotificationSetting.get_or_default(
                     player_id=self.initial_black_id, type='before_game_time', league=league)
                 old_black_setting.save()
-
+                
+            #set players available if game gets scheduled and either of them is set unavailable,
+            #do not set a player with a red card available, though.
+            if not SeasonPlayer.objects.filter(Q(player__id=self.white_id) & Q(season=self.get_round().season) & Q(games_missed__gte=2)).exists():
+                PlayerAvailability.objects.filter(
+                    (Q(player__id=self.white_id) & Q(round=self.get_round())
```
lakinwecker, line 1592: When using a `.filter()` specying multiple attributes to filter on already uses an `AND`. So that means that:
```python
.filter(Q(foo=X)&Q(bar=Y))
```
Is equivalent to:
```python
.filter(foo=X, bar=Y)
```

lakinwecker: You only need to use the `Q` objects for more complicated logic, like `OR` or mixing `OR` with `AND`

heltour/tournament/models.py
```diff
@@ -1583,7 +1584,19 @@ def save(self, *args, **kwargs):
...
+                  )).update(is_available=True)
+            
+            if not SeasonPlayer.objects.filter(Q(player__id=self.black_id) & Q(season=self.get_round().season) & Q(games_missed__gte=2)).exists():
+                PlayerAvailability.objects.filter(
+                    (Q(player__id=self.black_id) & Q(round=self.get_round())
```
lakinwecker, line 1597: I feel these two ifs can be quickly refactored to a single function and called twice, once for black and once for white. 

```python
def update_available_upon_scheduling(player_id):
    if not SeasonPlayer.objects.filter(player__id=player_id, season=self.get_round().season, games_missed__gte=2).exists():
        PlayerAvailability.objects.filter(player__id=player_id, round=self.get_round()).update(is_available=True)
```

lakinwecker (changes requested):
Some minor changes.

heltour/tournament/models.py
hopffgam: done

mathgeek:
Should red card players be blocked from scheduling? Or are they already and this is just a sanity check?

hopffgam:
> Should red card players be blocked from scheduling? Or are they already and this is just a sanity check?

Good question, difficult to answer. First of all this should only be relevant during the round where the player recieves the red card. If the red card is received for a no-show, I think there is nothing to do. However, if the red card is recieved for a late contact, this question is relevant: Ideally in the team league (where we have alternates) the red card players should really be blocked from scheduling. When they are late contact an alternate search will start anyway. At the moment they can schedule but not stop their alternate search, which is at least confusing for their opponents who might not be aware of the red card.
On the other hand in the individual leagues (currently lonewolf and chess960, where we do not have alternates) the lesser evil seems to be to let those players still play - and thus schedule - their games if their opponents agree. Those opponents can claim the win already today, but most of them prefer to play a game so imo it makes sense to let the red carded player play their game and only withdraw them afterwards. 

To get to the ideal solution the challenge would be to implement this different withdrawal behavior in a way that is not weird. Not sure if linking it to the league.competitor_type='team' is a good approach. In addition, what would need to be clarified is how chesster would have to change. Chesster should not just ignore a scheduling message that affects a red carded player but provide some sort of useful feedback and I am not sure if chesster would need to be changed for that and how complex that would be.

mathgeek:
> > Should red card players be blocked from scheduling? Or are they already and this is just a sanity check?
> 
> Good question, difficult to answer. First of all this should only be relevant during the round where the player recieves the red card. If the red card is received for a no-show, I think there is nothing to do. However, if the red card is recieved for a late contact, this question is relevant: Ideally in the team league (where we have alternates) the red card players should really be blocked from scheduling. When they are late contact an alternate search will start anyway. At the moment they can schedule but not stop their alternate search, which is at least confusing for their opponents who might not be aware of the red card. On the other hand in the individual leagues (currently lonewolf and chess960, where we do not have alternates) the lesser evil seems to be to let those players still play - and thus schedule - their games if their opponents agree. Those opponents can claim the win already today, but most of them prefer to play a game so imo it makes sense to let the red carded player play their game and only withdraw them afterwards.
> 
> To get to the ideal solution the challenge would be to implement this different withdrawal behavior in a way that is not weird. Not sure if linking it to the league.competitor_type='team' is a good approach. In addition, what would need to be clarified is how chesster would have to change. Chesster should not just ignore a scheduling message that affects a red carded player but provide some sort of useful feedback and I am not sure if chesster would need to be changed for that and how complex that would be.

Maybe something to open a future issue for then

lakinwecker (approved):
Looks good. Merging

PR #459, opened 2022-09-17
fixes #408 season player comments shown in edit rosters view

### LINKED ISSUE #408: Show SeasonPlayer comments on the Edit Rosters view
Currently, only Player comments are shown in the call-out views on players. We use the comments to indicate that a player should not be played in alternate queues again. It would be more logical to have these as SeasonPlayer comments, as they really only affect the current season. However, we have to use Player comments, as the SeasonPlayer comments are not shown in Edit Rosters

hopffgam:
fixes #408 Season player comments are now shown in the player callout on the "edit rosters" view


lakinwecker (approved):
LGTM

PR #460, opened 2022-09-17
409 registration form

### LINKED ISSUE #409: Setting availability with Change Registration form
In the registration form you can provide weeks you are unavailable. However, when you do that on the registration form after your registration had been approved already, this will only switch weeks from available to unavailable. Changing a week from unavailable to available will not be registered in the database.

Instead of fixing that particular issue I would be in favor of removing the possibility to set availabilities in the registration form after a registration had been approved already. Once a registration is approved players can use Edit Availability instead.

### LINKED ISSUE #447: add bolded instruction to registration form re: non-provisional classical
on our registration page, I think we should have in bold:

**Applications are only processed _after_ playing enough classical games to establish a non-provisional rating.  If there is a question mark next to your classical rating, your application will not be processed.**

hopffgam:
fixes #409 fixes #447 

only allow to specify availability on registration form if player has not been accepted for the season already (as then they can and should use "edit availability" instead of changing their registration.

Help text in the registration forms extended to make more clear that players need to get an established rating before registrations are processed. The helptext in lonewolf registrations mentions exceptions that might apply for lonewolf.

### DIFF
--- a/heltour/tournament/forms.py
+++ b/heltour/tournament/forms.py
@@ -44,6 +44,14 @@ class RegistrationForm(forms.ModelForm):
         self.fields['classical_rating'] = forms.IntegerField(required=True, label=_(
             'Your Lichess %s Rating' % rating_type))
 
+        help_text_provisional = 'Applications are only reviewed if you have an established rating. \
+            If your rating is still provisional (there is a question mark next to your rating), you can already register. \
+            But we will only review your registration once your rating is established.'
+        
+        if league.competitor_type != 'team':
+            help_text_provisional = 'Applications are only reviewed if you have an established rating. \
+            If there is a question mark next to your rating, your application will not be reviewed. \
+            Exceptions may apply if you have played in our league before. Please check the league rules.'
         # 20 games
         self.fields['has_played_20_games'] = forms.TypedChoiceField(required=True,
                                                                     choices=YES_NO_OPTIONS,
@@ -52,7 +60,7 @@ class RegistrationForm(forms.ModelForm):
                                                                     label=_(
                                                                         'Is your %s rating established (not provisional)?' % rating_type.lower()),
                                                                     help_text=_(
-                                                                        'If it is provisional, it must be established ASAP by playing more games.'), )
+                                                                        help_text_provisional), )
 
         # In slack
         self.fields['already_in_slack_group'] = forms.TypedChoiceField(required=True, label=_(
@@ -149,7 +157,8 @@ class RegistrationForm(forms.ModelForm):
                                                                     label=_(
                                                                         'Are you interested in being an alternate or a full time player?'),
                                                                     help_text=_(
-                                                                        'Players are put into teams on a first come first served basis, you may be an alternate even if you request to be a full time player.'))
+                                                                        'Players are put into teams on a first come first served basis, based on registration date. \
+                                                                        You may be an alternate even if you request to be a full time player.'))
         else:
             del self.fields['alternate_preference']
 
--- a/heltour/tournament/forms.py
+++ b/heltour/tournament/forms.py
@@ -36,6 +36,23 @@ class RegistrationForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         self.season = kwargs.pop('season')
+        
+        username = kwargs.pop('user')
+        
+        #try to find a SeasonPlayer object for the current season.
+        #if SeasonPlayer object already exists for the current season we know the player got accepted before 
+        #and we do not have to display the availability fields in the registration form.
+        try:
+            #the lonewolf seasons are organized in section groups. Registrations are always linked to the lonewolf open section
+            #but players might be assigned to the u1800 section. Hence we have to identify all seasons in the same section group.
+            #On the other hand, team league and 960 do not have sections and we have to check directly the season.
+            seasonsInGroup = Season.objects.filter(
+                section__in=Section.objects.filter(
+                    section_group=Section.objects.filter(season=self.season).first().section_group))
+        except:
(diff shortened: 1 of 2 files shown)


heltour/tournament/forms.py
```diff
@@ -36,6 +36,23 @@ class Meta:
...
+        try:
+            #the lonewolf seasons are organized in section groups. Registrations are always linked to the lonewolf open section
+            #but players might be assigned to the u1800 section. Hence we have to identify all seasons in the same section group.
+            #On the other hand, team league and 960 do not have sections and we have to check directly the season.
+            seasonsInGroup = Season.objects.filter(
+                section__in=Section.objects.filter(
+                    section_group=Section.objects.filter(season=self.season).first().section_group))
+        except:
+            seasonsInGroup = Season.objects.filter(name=self.season.name)
+        already_accepted = SeasonPlayer.objects.filter(season__in=seasonsInGroup, player__lichess_username=username).exists()
```
hopffgam, line 54: I did not find a smarter way to write this. Problem is that for lonewolf the seasons are linked with each other via Section and SectionGroup, but team league and chess960 have no section. Pretty sure there is a better way than try/except but the checks that I have tried (self.season.section is None) resulted in exceptions for the team league.

heltour/tournament/forms.py
```diff
@@ -44,6 +61,14 @@ def __init__(self, *args, **kwargs):
         self.fields['classical_rating'] = forms.IntegerField(required=True, label=_(
             'Your Lichess %s Rating' % rating_type))
 
+        help_text_provisional = 'Applications are only reviewed if you have an established rating. \
```
lakinwecker, line 64: I feel that we should review the text here. I think it can be simplified. 

Team:
> You may apply with a provisional rating, but your application will only be reviewed once your rating is established.

Not-team:
> You may apply with a provisional rating, but your application will only be reviewed once your rating is established unless there are league-specific rules that override this.

heltour/tournament/forms.py
```diff
@@ -36,6 +36,23 @@ class Meta:
...
+        #if SeasonPlayer object already exists for the current season we know the player got accepted before 
+        #and we do not have to display the availability fields in the registration form.
+        try:
+            #the lonewolf seasons are organized in section groups. Registrations are always linked to the lonewolf open section
+            #but players might be assigned to the u1800 section. Hence we have to identify all seasons in the same section group.
+            #On the other hand, team league and 960 do not have sections and we have to check directly the season.
+            seasonsInGroup = Season.objects.filter(
+                section__in=Section.objects.filter(
+                    section_group=Section.objects.filter(season=self.season).first().section_group))
+        except:
```
lakinwecker, line 52: Bare excepts are dangerous. Catch only the exceptions that are necessary.

heltour/tournament/forms.py
lakinwecker: You could use the same `Q` objects we did previously.  What, exactly, are the exceptions that are throw in the first case?

heltour/tournament/forms.py
lakinwecker: In general, python uses snake case and I think that heltour does too. We should use the name `seasons_in_group`

heltour/tournament/forms.py
lakinwecker: I'm unclear why this wouldn't work:
```python
potential_sections = Section.objects.filter(
    season=self.season
)

seasonsInGroup = Season.objects.filter(
    Q(section__in=potential_sections) | Q(name=self.season.name)
)

already_accepted = SeasonPlayer.objects.filter(season__in=seasonsInGroup, player__lichess_username=username).exists()
```

lakinwecker (changes requested):
Minor feedback and suggestions.

heltour/tournament/forms.py
hopffgam: Argh, Season has a method that returns exactly what I want (facepalm), see recent commit.

mathgeek:
This looks a lot more straightforward now

heltour/tournament/templates/tournament/register.html
```diff
@@ -24,6 +24,13 @@ <h3>{{ registration_season.section_group_name }} Registration</h3>
                             {{ registration_season.start_date|date_or_q }}{% if registration_season.end_date %} and ends
                                 {{ registration_season.end_date|date_or_q }}{% endif %}.</p>
                     {% endif %}
+                    <p>You are registering as <strong>{{ user }}</strong>. 
+                    <br>If you want to change the account you register with, you can logout <a href="/team4545/player/">here</a>.
+                    <br>As per our 
+                    <a href="/team4545/document/terms-of-service/">Terms of Service</a> once you 
+                    participate in our league with this account you will always need to use it.
+                    
+                    </p>
```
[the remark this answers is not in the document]
mathgeek: Not sure what would be done with older accounts where this was implicit in the signup (with users entering their username basically being the confirmation that they want to sign up with that account?). Also IF we decide to add this and track it it should be tracked by the date the user agreed to this and only updated one time.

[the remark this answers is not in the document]
mathgeek: Actually I realized as soon as I submitted that comment that this object is probably season-specific right? That would temper a lot of the concerns I had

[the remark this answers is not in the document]
hopffgam: Registrations are season specific, but those things would - like the dates for ToS consent - have to be stored on the player objects. However, my suggestion would be that if we want to make players confirming that, we create a separate issue, as such functionality is probably not urgent, while I would like to get some of the other features in this PR deplyoed.

PR #464, opened 2022-09-22
fixes #418 no update of player data when account is closed

### LINKED ISSUE #418: Closed accounts get rating deleted
This causes issues all over the place as many calls expect that the rating of a player exist as integer value

hopffgam:
fixes #418 
- the JSON object that stores account data is not updated anymore after an account got closed.
- If this JSON object does not exist at all, a rating of 0 is returned to screens that display the player rating. This should only happen rarely for new players, but is relevant for those players where we lost the data. This fixes several admin screens that ordered players by rating, particularly the team editing screen.

### DIFF
--- a/heltour/tournament/models.py
+++ b/heltour/tournament/models.py
@@ -789,15 +789,17 @@ class Player(_BaseModel):
                                                        new_value=self.account_status)
 
     def update_profile(self, user_meta):
-        self.profile = user_meta
         classical = user_meta.get('perfs', {}).get('classical')
         if classical is not None:
             self.rating = classical['rating']
             self.games_played = classical['games']
-        
         is_closed = user_meta.get('disabled', False)
         is_tosViolation = user_meta.get('tosViolation', False)
         self.account_status = 'closed' if is_closed else 'tos_violation' if is_tosViolation else 'normal'
+        
+        # profile is used to get rating data which should not be updated anymore once an account is closed.
+        if not is_closed:
+            self.profile = user_meta
         self.save()
 
     @classmethod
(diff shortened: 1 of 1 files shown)


mathgeek (approved):

heltour/tournament/models.py
```diff
@@ -827,7 +829,10 @@ def is_available_for(self, round_):
     def rating_for(self, league):
         if league:
             if self.profile is None:
-                return None
+                # some admin screens cannot handle a None rating, so we return 0 instead.
+                # self.profile is only None if the player profile has never been downloaded
+                # from lichess or the account had already been closed at that first download.
+                return 0
```
lakinwecker, line 835: Shoudl we should return `1500`?. The logic is: ff we don't know the rating, the assumption is the default rating for lichess with a high RD, not 0, which is a completely invalid rating.

However, that means that in these views, we can't distinguish between something without a rating and and something with a legit default rating.

lakinwecker (changes requested):
Minor question about the value, happy to merge it either way, I just want to ensure that we make a conscious decision about the value.

heltour/tournament/models.py
mathgeek: I like 0 for the reason that it is an invalid rating, not only is 1500 the default rating it’s also a legitimate rating in general. It makes it very obvious that these users are a special case (and in the case of sort would be at the bottom). Plus for some of these users we already know that their last known rating was an established one, we’ve just since had the rating blanked out

hopffgam: I also prefer to return a completely invalid rating (i.e. 0) that still fulfills the technical expectation of the callers of the function, i.e. to receive a numeric value:
This value should only be returned for players who closed their account while we had the bug we are fixing in  Player.update_profile. If the rest of this fix works, this line is really only to prevent admin screens for older seasons from crashing, where players are listed who closed their account in the last 1.5 years or so. Nevertheless, even if those cases show up in the future, at least it is easy to see that something is wrong with the rating.

lakinwecker: Works for me.

PR #466, opened 2022-09-24
withdrawal handling in team league

### LINKED ISSUE #465: Withdrawal handling for team league
In team league the player withdrawals have to be handled by moderators manually. Ideally the following things should happen automatically:
1. The player is set unavailable for all future rounds of the season
2. A comment is added to the SeasonPlayer object about when the player withdrew
3. If the withdrawal is before team creation, the SeasonPlayer object could be set to is_active=False so that the player does not even appear in the Edit Teams screen. But that is low priority.
4. If the player is an alternate, the SeasonPlayer object could be set to is_active=False so that the player is removed from the alternate lists.
5. If the player is member of a team a message from chesster to the captains channel could be sent (registered as #448)

hopffgam:
fixes #465
when a player withdraws for team league they are
- set unavailable for future rounds
- if they are an alternate the SeasonPlayer is set inactive
- if teams have not been created yet the SeasonPlayer is set inactive
- A comment about the withdrawal is added to the SeasonPlayer
- sending message to captains channel is skipped. This is handled in issue 448

### DIFF
--- a/heltour/tournament/models.py
+++ b/heltour/tournament/models.py
@@ -990,7 +990,13 @@ class PlayerWithdrawal(_BaseModel):
                                          type='full-point-pairing-bye')
                 pairing.delete()
 
+    def perform_team_season_withdrawal(self):
+        sp = SeasonPlayer.objects.get(season=self.round.season, player=self.player)
+        sp.withdraw_from_team_season(self.round)
+
     def save(self, *args, **kwargs):
+        if self.round.season.league.is_team_league():
+            self.perform_team_season_withdrawal()
         super(PlayerWithdrawal, self).save(*args, **kwargs)
         if self.round.publish_pairings and not self.round.is_completed:
             self.perform_withdrawal()
@@ -1814,11 +1820,12 @@ class SeasonPlayer(_BaseModel):
         self.initial_player_id = self.player_id
         self.initial_games_missed = self.games_missed
         
-    def _set_unavailable_for_season(self):
+    def _set_unavailable_for_season(self, skip_current=False):
         rounds = Round.objects.filter(season=self.season, is_completed=False)
         for r in rounds:
-            PlayerAvailability.objects.update_or_create(round=r, player=self.player,
-                                                    defaults={'is_available': False})
+            if not (skip_current and r.publish_pairings):
+                PlayerAvailability.objects.update_or_create(round=r, player=self.player,
+                                                        defaults={'is_available': False})
             
     def has_scheduled_game_in_round(self, round):
         pairingModel = TeamPlayerPairing.objects.filter(team_pairing__round=round)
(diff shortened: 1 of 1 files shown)


heltour/tournament/models.py
```diff
@@ -1872,6 +1879,15 @@ def seed_rating_display(self, league=None):
...
             return self.player.rating_for(league)
+    
+    def withdraw_from_team_season(self, round):
+        #we should only set players inactive here if they are not member of a team,
+        #i.e. if they are either alternates or we are in the pre-season before team creation.
+        if not TeamMember.objects.filter(player=self.player, team__season=round.season).exists():
+            self.is_active = False
+            self.save()
+        self._set_unavailable_for_season(skip_current=True)
+        add_system_comment(self, 'player withdrawn for %s'%round)
```
mathgeek, line 1890: Might want to make this “from round X” - also if this is in preseason what will the round be? Mostly for my ignorance here

mathgeek:
One minor comment otherwise looks good, might want Lakin to double check though

heltour/tournament/models.py
hopffgam: Withdrawals during pre-season are registered for round 1 (I have just double-checked that). The %s here resolves to a string "Season 31 - Round 1" for example. Maybe "for" and "from" are both not ideal but we could just make it "player withdrawn: %s"

mathgeek: No this is perfectly fine with me then

heltour/tournament/models.py
```diff
@@ -1814,11 +1820,12 @@ def __init__(self, *args, **kwargs):
         self.initial_player_id = self.player_id
         self.initial_games_missed = self.games_missed
         
-    def _set_unavailable_for_season(self):
+    def _set_unavailable_for_season(self, skip_current=False):
         rounds = Round.objects.filter(season=self.season, is_completed=False)
         for r in rounds:
-            PlayerAvailability.objects.update_or_create(round=r, player=self.player,
-                                                    defaults={'is_available': False})
+            if not (skip_current and r.publish_pairings):
```
lakinwecker, line 1826: I don't know why, but I find this if statement more confusing than if we inverted it and used `continue`

```python
    if skip_current and r.publish_pairings:
        continue
    PlayerAvailability.objects.update_or_create(round=r, player=self.player,
                                                 defaults={'is_available': False})


        
```

heltour/tournament/models.py
```diff
@@ -1872,6 +1879,15 @@ def seed_rating_display(self, league=None):
             if league is None:
                 league = self.season.league
             return self.player.rating_for(league)
+    
+    def withdraw_from_team_season(self, round):
```
lakinwecker, line 1883: I like the idea of making this be an `@classmethod` so it can be called like I suggested in the other comment, which makes this signature read:
```python
@classmethod
def withdraw_from_season(cls, round, player):
    if not TeamMember.objects.filter(player=self.player, team__season=round.season).exists():
        cls.objects.filter(season=self.round.season, player=self.player).update(is_active=False)
        self.is_active = False
        self.save()
    cls._set_unavailable_for_season(player, skip_current=True)
    add_system_comment(self, 'player withdrawn: %s'%round)
```

Note: this makes it so that you have to update `_set_unavailable_for_season` to be callable either as a `@classmethod` or as an instance method. Which shouldn't be too hard. Maybe just make a new one that's the `@classmethod` with a different name and update the current one to use it.

hopffgam: as we need the `SeasonPlayer` for add_system_comment as well, I am getting the correct `SeasonPlayer sp` and then call `sp._set_unavailable_for_season` instead of making another `@classmethod`

heltour/tournament/models.py
hopffgam: done

lakinwecker (approved):

PR #477, opened 2022-12-28
fixes #449 improved accessibility of availability screen

### LINKED ISSUE #449: Availability slots in availability form not read by screen readers
The toggles in the availability form are not read by screen readers. Blind users of our website don't know if a toggle is set to "available" or to "unavailable".

hopffgam:
fixes #449 by adding an aria-label that is linked to the current setting of the availability toggles to present hints to blind users. Not a great solution imo but short term the best I can provide. Long term solution might be to replace the bootstrap toggles altogether or to improve the implementation of these hints.

### DIFF
--- a/heltour/tournament/templates/tournament/availability.html
+++ b/heltour/tournament/templates/tournament/availability.html
@@ -58,20 +58,18 @@
                                     <td>{{ r.start_date|date_or_q:'%b %-d' }} - {{ r.end_date|date_or_q:'%b %-d' }}</td>
                                     {% for p, is_av, is_immutable in av_list %}
                                         <td>
-                                            {% with id="{{ r.number }}_{{ p.lichess_username }}" %}
-                                                <input id="availability-checkbox-{{ id }}"
-                                                       type="checkbox" name="av_r{{ id }}"
-                                                       data-toggle="toggle" data-on="Unavailable" data-off="Available"
-                                                       data-onstyle="default" data-offstyle="success" data-size="small"
-                                                       onchange="availabilityChanged(this, document.getElementById('availability-checkbox-label-{{ r.number }}_{{ p.lichess_username }}'))"
-                                                       {% if not is_av %}checked{% endif %}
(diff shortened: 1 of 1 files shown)


heltour/tournament/templates/tournament/availability.html
```diff
@@ -13,6 +13,17 @@
...
+
+    <script>
+    function availabilityChanged(checkbox, label) {
+      if (checkbox.checked) {
+          label.setAttribute("aria-label","Current setting: unavailable - press save to submit");
+      } else {
+          label.setAttribute("aria-label", "Current setting: available - press save to submit");
+      }
+    }
+</script>
```
lakinwecker, line 25: indentation is inconsistent here.

heltour/tournament/templates/tournament/availability.html
```diff
@@ -47,12 +58,18 @@ <h3>Edit Availability</h3>
                                     <td>{{ r.start_date|date_or_q:'%b %-d' }} - {{ r.end_date|date_or_q:'%b %-d' }}</td>
                                     {% for p, is_av, is_immutable in av_list %}
                                         <td>
-                                            <input type="checkbox" name="av_r{{ r.number }}_{{ p.lichess_username }}"
+                                            <input id="availability-checkbox-{{ r.number }}_{{ p.lichess_username }}"
```
lakinwecker, line 61: I need to double check, but you might be able to use the django `{% with .. %}` tag to make this code a bit nicer: https://docs.djangoproject.com/en/4.1/ref/templates/builtins/#with

lakinwecker: The idea is that I'm hoping something like this works:
```django
{% with id="{{ r.number }}_{{ p.lichess_username }}" %}
<input id="availability-checkbox-{{ id }}"
{% endwith %}
```

hopffgam: Thanks, I will try that.

PR #481, opened 2023-02-11
round mean in lw stats

### LINKED ISSUE #478: mean on stats page not rounded
the mean for the rating distribution on the stats page of (at least) lonewolf and chess960 is not rounded.
see e.g. https://www.lichess4545.com/lonewolf/season/20/stats/ where we can currently see that the means are 1984.0102040816328 and 1963.7202797202797 for active and all players respectively. these should probably be rounded to 1984 and 1964. maybe 1984.0 and 1963.7 if we are feeling generous.

as an aside, the stats page has several other numbers that could/should probably be rounded more aggressively.

i am marking this "good first issue" because there's no way this is hard to fix.

glbert-does:
closes #478


lakinwecker:
Looks good to me.

PR #482, opened 2023-02-15
allow provisional players for team

### LINKED ISSUE #473: Team creation: accepted players with provisional rating should be placed on rosters
If a player has a provisional rating at the time of team creation but got accepted for the season it means that their rating was established at the time of approval and got provisional again. We allow those players to participate, and thus they should be treated the same way as players with established rating and should be assigned to a team or alternate board according to their preference and registration date.

glbert-does:
fixes #473. i have confirmed that this works as the team mods want in a small test season.
i have also decided to leave provisional players in red_players, that way it is still possible to check for registrations that were accepted by mistake.

### DIFF
--- a/heltour/tournament/teamgen.py
+++ b/heltour/tournament/teamgen.py
@@ -147,7 +147,7 @@ def flatten(lst):
 def make_league(playerdata, boards, balance):
     players = []
     for player in playerdata:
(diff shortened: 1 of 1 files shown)


lakinwecker (approved):
LGTM

PR #484, opened 2023-02-15
add 3rd alternate option at team registration and team creation

### LINKED ISSUE #444: 3 options for alternate choice
1. Full-time
2. Alternate
3. Either is fine for me

This would need to be implemented (at least) in (1) the registration form, and (2) teamgen.py, and optionally the manage_players page as well.
Note that the mods could already work with it being implemented in the registration form, which would allow us to manually set undecided players according to what is needed currently.

glbert-does:
closes #444
tested on a small test season only


heltour/tournament/teamgen.py
```diff
@@ -19,13 +19,14 @@ class Player:
     board = None
     req_met = False
 
-    def __init__(self, name, rating, friends, avoid, date, alt, previous_season_alt):
+    def __init__(self, name, rating, friends, avoid, date, alt, altfine, previous_season_alt):
```
lakinwecker, line 22: For consistency, this should also be `alt_fine`

lakinwecker (changes requested):
Minor nit.

PR #485, opened 2023-02-17
API get_season_games accepts a parameter include_unplayed

### LINKED ISSUE #472: API get_season_games should export unplayed results as well
currently here:
https://github.com/Lichess4545/heltour/blob/c0457a412cf8b8af8b2d7d1667e2cd6837641e1b/heltour/tournament/api.py#L573-L588
only games with game_id set are exported; that is unplayed forfeits or scheduling draws are not included.
imo they should be, or maybe there should be an api call get_season_results that includes unplayed resuls as well.

glbert-does:
include_unplayed=true adds unplayed pairings to the json output.
include_unplayed defaults to false, so the default behaviour is unchanged.

closes #472.

### DIFF
--- a/heltour/tournament/api.py
+++ b/heltour/tournament/api.py
@@ -571,22 +571,22 @@ def get_season_games(request):
                                            'teamplayerpairing__team_pairing__white_team',
                                            'teamplayerpairing__team_pairing__black_team'):
             game_id = get_gameid_from_gamelink(p.game_link)
-            r = p.get_round()
-            g = {
-                'league': s.league.name,
-                'season': s.name,
-                'round': r.number if r else None,
-                'game_id': game_id if game_id else None,
-                'white': p.white.lichess_username if p.white else None,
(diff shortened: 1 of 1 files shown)


heltour/tournament/api.py
```diff
@@ -570,13 +571,13 @@ def get_season_games(request):
...
             game_id = get_gameid_from_gamelink(p.game_link)
-            if game_id:
+            if game_id or include_unplayed:
                 r = p.get_round()
                 g = {
                     'league': s.league.name,
                     'season': s.name,
                     'round': r.number if r else None,
-                    'game_id': game_id,
+                    'game_id': game_id if game_id else None,
```
lakinwecker, line 580: I am guessing that when `game_id` is "falsey", then it's probably actually set to `None` already, rendering this line redundant.

lakinwecker (changes requested):
Minor nit

PR #493, opened 2023-03-12
send an email on rejection

### LINKED ISSUE #260: Add templates that generate email when rejecting registrations
When we reject registrations, we do so for a number of different reasons:

1. Duplicate registrations.
2. Rejecting due to ineligibility to participate in the league and there is no hope of becoming eligible.
3. Rejecting due to lack of games played (provisional rating).

It would be good to record the reason why (we currently do so via comment), and additionally it would be good if we could select from a set of pre-made email templates for the second 2 options and have an email be sent to registration.

glbert-does:
closes #260

### DIFF
--- a/heltour/tournament/admin.py
+++ b/heltour/tournament/admin.py
@@ -2115,7 +2115,6 @@ class RegistrationAdmin(_BaseAdmin):
                         reg.status_changed_date = timezone.now()
                         reg.save()
 
-
                     self.message_user(request,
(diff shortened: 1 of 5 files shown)


lakinwecker (commented):
This looks good from a review, but I'd like to test it before I merge it. I'll try and find time this week.

lakinwecker (approved):
LGTM

PR #494, opened 2023-03-13
Remove several registration fields that are not required anymore

### LINKED ISSUE #488: shorten registration form
* rating can easily be fetched from lichess and is not needed.

* we know whether the account of a player is linked. asking if the player is on the slack group is somewhat unnecessary.

* we know whether someone was an alt in their last season or not. [Team league only]

for all 3 of the above, i believe that the information given by the user is largely not used anymore anyway. make sure that is the case. if it is still used, look into fixing that. if it is not used, the item can be removed from the registration form.

### LINKED ISSUE #489: shorten team registration form (conditionally)
after a season has started, as far as i can see the following items are not necessary:
* alternate preference
* friends list
* avoid list

if possible, remove them conditionally when the season has started and the first pairings were created.

### LINKED ISSUE #492: in registration form, don't ask about provisional rating, but inform instead
we don't need to ask whether a player's rating is provisional, we can easily get that information from lichess. instead, we should inform them in case their rating is provisional.

this could also be conditional on whether the league allows provisional ratings for veteran players, like lw and chess960 do.

### LINKED ISSUE #539: when registering, alts should not have to confirm that they can commit to one game per week
> Are you able to commit to 1 long time control game (45+45 currently) of classical chess on Lichess.org per week?

i think at minimum the text of the question should reflect that. even better, maybe the text could change for people who picked alt, or the question could be skipped for alts.

sometimes people get spooked by this, so i will label it as "medium". wouldn't want to lose potential players over what is essentially not a great phrasing by us.

i will ask the team mods if the have some preference for a different phrasing, as the minimal effort solution.

glbert-does:
when ready, this will close #488

* rating (is fetched from the site via the api)
* peak rating (could be fetched from the site and maybe still is; but this has not been used in ages anyway)
* previous season alt (we know who was an alt last season – i am not completely sure everything works correctly here though)
* already in slack group (we know that, it's players who are linked)
* slack_username (as far as i know, this hasn't been used in years)

i removed "expected rating" at the same time. this hasn't been used in ages, and cannot even be used with the current team gen algorithm. frankly, i am not certain it ever made sense to use it.

edit: also closes #489, closes #492, closes #539


heltour/tournament/migrations/0021_auto_20160724_1855.py
```diff
@@ -20,7 +20,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='registration',
             name='already_in_slack_group',
-            field=models.BooleanField(),
```
lakinwecker, line 23: Altering old migrations has no effect on the live database, what was the intent with this change?

lakinwecker (changes requested):
I haven't finished a full review, but I thought I would send my first comments anyways. I'll want to do a deeper review and test this a bit, which don't have time for today. But first let's deal with changing the old migrations. What was the intent for those?

heltour/tournament/migrations/0021_auto_20160724_1855.py
glbert-does: the intent was simply to make the new migrations revertible. as far as i can tell, that's the only thing it does, and i didn't see a downside to it. i don't mind removing the changes to the old migrations from the pull request.

lakinwecker: I am pretty sure there are ways to make the new migrations invertible that don't include modifying old migrations. The problem with modifying old migrations is that our live database will not be in sync with this change, which means that new migrations may assume they can do things that they can't and it will succeed in development, but not on prod. Let's remove the changes to old migrations and if you want the new migrations to be revertible, we can work through that.

glbert-does:
now also closes #489

glbert-does:
now also closes #492

PR #499, opened 2023-04-02
Fix 500 errors from bad requests

cyanfish:


lakinwecker:
I'll let you merge it. I'll deploy afterwards

cyanfish:
Thanks. There's not a security issue or anything like that, but after getting 200 email notifications when someone was pentesting I figured it's worth blocking :)

PR #500, opened 2023-04-09
Use settings instead of hardcoding

glbert-does:
amongst many, many other places, "lichess.org" is hardcoded into the api_worker. if i change my settings to lichess.dev, and try to login via a lichess.dev user that does not exist on lichess.org, this will lead to an error in the api_worker and the login fails.

there are likely other places where reading the settings would make testing in other places than .org easier, but this one fixes my immediate problem.


heltour/settings_default.py
```diff
@@ -301,10 +301,10 @@
 CHESSTER_USER_ID = 'U0VCPUT7T'
 
 LICHESS_DOMAIN = 'https://lichess.org/'
-LICHESS_OAUTH_ACCOUNT_URL = 'https://lichess.org/api/account'
-LICHESS_OAUTH_EMAIL_URL = 'https://lichess.org/api/email'
-LICHESS_OAUTH_AUTHORIZE_URL = 'https://lichess.org/oauth'
-LICHESS_OAUTH_TOKEN_URL = 'https://lichess.org/api/token'
```
lakinwecker, line 307: I have a very minor preference for using format strings for stuff like this, but it's not enough to prevent merging this.

lakinwecker (approved):
LGTM

heltour/settings_default.py
glbert-does: i will try to remember in the future.

PR #502, opened 2023-05-11
be more forceful about using slack instead of the contact form

glbert-does:
asking people with linked slack account to use slack instead in red letters. also, for logged in users, fill the username into the form as initial value.

### DIFF
--- a/heltour/tournament/views.py
+++ b/heltour/tournament/views.py
@@ -1439,6 +1439,10 @@ class ContactView(LeagueView):
         leagues = [self.league] + list(
             League.objects.filter(is_active=True).order_by('display_order').exclude(
                 pk=self.league.pk))
+
+        player = Player.get_or_create(self.request.user.username)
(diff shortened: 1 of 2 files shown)


heltour/tournament/views.py
```diff
@@ -1439,6 +1439,10 @@ def view(self, post=False):
         leagues = [self.league] + list(
             League.objects.filter(is_active=True).order_by('display_order').exclude(
                 pk=self.league.pk))
+
+        player = Player.get_or_create(self.request.user.username)
+        slack_linked = bool(player.slack_user_id)
```
lakinwecker, line 1444: You don't need to explicitly cast this, although it's not harmful. You could just pass this value into the context directly and the if statement would still work, and it would be considered slightly more pythonic, I think.

lakinwecker (changes requested):
Minor comment about extra unneeded line. It's not strictly necessary, but figured it was worth requesting.

PR #506, opened 2023-05-30
less hardcoding of lichess.org

glbert-does:
for testing i want to use games on lichess.dev, don't hardcode the game_link_validator to lichess.org but use domain from settings instead.


lakinwecker (approved):
LGTM

PR #507, opened 2023-06-03
validate pending registrations daily

### LINKED ISSUE #200: Track/notify for when registered player ratings become established
And automatically trigger re-validation.

glbert-does:
once a day, validate pending registrations. this closes #200, in a manner of speaking. instead of updating the rating and only then triggering validation on players with established rating, it simply triggers validation of all pending registrations. i am not certain how big the difference is, since as far as i can tell, a validation will force an update of the rating anyway.

running this every second day would be perfectly ok as well.

### DIFF
--- a/heltour/settings_default.py
+++ b/heltour/settings_default.py
@@ -216,6 +216,11 @@ CELERYBEAT_SCHEDULE = {
         'schedule': timedelta(minutes=1),
         'args': ()
     },
+    'validate_pending_registrations': {
+        'task': 'heltour.tournament.tasks.validate_pending_registrations',
+        'schedule': timedelta(days=1),
+        'args': ()
+    },
     'celery_is_up': {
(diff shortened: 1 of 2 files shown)


lakinwecker (approved):
I assume you have tested this? If so, then it looks good to me. If you have, then go ahead and merge, if you haven't, let me know and I'll test it before I merge.

glbert-does:
i did some tests – but they were probably not as exhaustive as they should be. i basically set it to run every minute and added one report of every useful type i could think of, checking that only regs were revalidated where it is useful, and that validated regs changed as a appropriate. so, in essence i tested it on maybe 20 regs.

i don't think i would feel comfortable to merge right now myself; i can run more tests in about a week or two. if someone else can do more tests before that, that would of course be fine for me. it would probably be helpful for mods to have this working during pre-season.

PR #523, opened 2023-11-01
Start games automatically

### LINKED ISSUE #491: Start games automatically
look into using https://lichess.org/api#tag/Bulk-pairings/operation/bulkPairingCreate to create games and start clocks.

here is how i personally currently imagine it:
* once a game is scheduled, both players need to confirm that it was scheduled correctly, maybe via a button on the site, or via chesster. maybe chesster should send a site link. if either player doesn't, they will have to still start the game manually.
* game starts a couple of minutes before the scheduled time (2-3 minutes maybe?), clock starts at the scheduled time.
* bulk pairings should of course create all the games that are supposed to start at a specified time.
* can even set "noClaimWin", which would enforce one of our rules.
* need to be careful about players creating their own challenge; definitely cannot create a game when the players already started on themselves, so would have to check for that first.

all of this is very low priority of course. but it would be quite fancy if we had something like this.

glbert-does:
closes #491, rebased from #527

with this pull request, we try to use the lichess bulk pairing api to create games and start clocks automatically at the scheduled time. here is how it is supposed to work:

* new league setting `start_games`. if this is set to true,
* players can confirm the time they have scheduled for.
* if the scheduled time changes, players need to confirm again.
* every */5 minutes scheduled games 30 seconds to 5:30 in the future are started (if both players confirmed that is), clocks are set to start in 6 minutes. lichess sends a message pointing to the game using the account that did the apicall.
* in practice this should mean that games scheduled for "nice" numbers like :00 or :30 are started five minutes before they are scheduled, and the clocks are started one minute after the game is scheduled.
* players cannot claim a win (yay!), but they can abort a game. which should help if an opponent does not show up. potentially, we could disallow early draws (move 30), but i don't think this would do much other than annoy players.

at least the following things are potential problems:

* **the account doing the bulk pairing needs a token with challenge:bulk scope**. i believe in our case this would be lichess.org/@/lichess4545 – i assume the token currently used might not have that yet.
* ~~if we do not have a valid token for a player, or the account is closed, lichess returns an error and lists the token(s) in question, without starting any games. i try to handle this, but it is rather not elegant. i assume there should be a better way than to use regexp on the error message~~. should work better now.
* ~~after 20 minutes, we used to check whether games have started by simply checking for a game id. this is not enough anymore, i now try to check in the move list whether more than one move was made.~~ at this point i have tested this a lot of times, and everything seems to be working as expected.
* the page to confirm scheduled times is rather ... bad. i did not put too much work into it, because i assume long term we will create a way to confirm times via slack.

i tried to test a lot of stuff, but it is difficult to get it anywhere close to what it would look like in prod.


glbert-does:
todo:
* the lichess response to a successful pairing could be used to set the `gameid`
* as mentioned, a way to confirm a scheduled time via slack 
* improve error handling
* potentially, if lichess rejects a token, we should invalidate it on our end

glbert-does:
most of the important todos are done. there's a couple nice-to-haves as well:
1. [ ] as mentioned, a way to confirm a scheduled time via slack
2. [ ] when a token was declined by lichess, it would be nice to force them to refresh it
3. [ ] possibly, whether a time was confirmed should be shown on the pairings page
4. [ ] potentially, it could be nice to have at least the time for starting the clocks be a league setting
5. [ ] maybe whether the clocks will be started at all could be a league setting too

heltour/tournament/tasks.py
```diff
@@ -292,6 +298,94 @@ def update_slack_users():
...
+           else: # no tokens left after deleting rejected tokens
+               result = None
+        except KeyError:
+            logger.exeption(f'[ERROR] could not parse error as json for {leaguename}:\n{e}')
+    # use lichess reply to set game ids
+    for game in league_games:
+        try:
+            for gameids in result['games']:
+                if gameids['white'] == game.white.lichess_username.lower() and \
+                   gameids['black'] == game.black.lichess_username.lower():
```
nicwest, line 338:
```suggestion
                if (gameids['white'] == game.white.lichess_username.lower() and
                   gameids['black'] == game.black.lichess_username.lower()):
```

heltour/tournament/tasks.py
```diff
@@ -292,6 +298,94 @@ def update_slack_users():
...
+                       game.game_link = get_gamelink_from_gameid(gameids['id'])
+                       game.save()
+                       signals.notify_players_game_started.send(sender=_start_league_games, \
+                                                                pairing=game, \
+                                                                gameid=gameids['id'])
```
nicwest, line 343:
```suggestion
                       signals.notify_players_game_started.send(sender=_start_league_games,
                                                                pairing=game,
                                                                gameid=gameids['id'])
```

heltour/tournament/tasks.py
```diff
@@ -292,6 +298,94 @@ def update_slack_users():
...
+
+
+@app.task()
+def start_games():
+    logger.info('[START] Checking for games to start.')
+    games_to_start = PlayerPairing.objects.filter(result='', game_link='', \
+            scheduled_time__lt=timezone.now() + timedelta(minutes=5, seconds=30), \
+        scheduled_time__gt=timezone.now() + timedelta(seconds=30), \
+            white_confirmed=True, black_confirmed=True) \
+            .exclude(white=None).exclude(black=None).select_related('white', 'black').nocache()
```
nicwest, line 357:
```suggestion
    games_to_start = PlayerPairing.objects.filter(
            result='', 
            game_link='', 
            scheduled_time__lt=timezone.now() + timedelta(minutes=5, seconds=30), 
            scheduled_time__gt=timezone.now() + timedelta(seconds=30), 
            white_confirmed=True,
            black_confirmed=True
       ).exclude(white=None).exclude(black=None).select_related('white', 'black').nocache()
```

heltour/tournament/tasks.py
```diff
@@ -292,6 +298,94 @@ def update_slack_users():
...
+    token_dict = {}
+    for game in games_to_start:
+        if (hasattr(game, 'loneplayerpairing')):
+            gameleague = game.loneplayerpairing.round.season.league
+        if (hasattr(game, 'teamplayerpairing')):
+            gameleague = game.teamplayerpairing.team_pairing.round.season.league
+        if gameleague is not None and gameleague.get_leaguesetting().start_games:
+            white_token = game.white.oauth_token.access_token
+            black_token = game.black.oauth_token.access_token
+            if (not white_token is None and not black_token is None and not game.white.oauth_token.is_expired() and not game.black.oauth_token.is_expired()):
```
nicwest, line 368: `not x is y` is less readable that `x is not y`

heltour/tournament/views.py
```diff
@@ -1801,6 +1801,54 @@ def view(self, post=False):
     def view_post(self):
         return self.view(post=True)
 
+class ConfirmScheduledTimeView(LeagueView, LoginRequiredMixin):
+    def view(self, post=False):
+        if not self.request.user.is_authenticated:
+            return redirect('by_league:league_home', self.league.tag)
```
nicwest, line 1807: do you need this if you have the `LoginRequiredMixin`?

heltour/tournament/views.py
```diff
@@ -1801,6 +1801,54 @@ def view(self, post=False):
     def view_post(self):
         return self.view(post=True)
 
+class ConfirmScheduledTimeView(LeagueView, LoginRequiredMixin):
```
nicwest, line 1804: as a general rule of thumb with multiple inheritance, you want to go mixins first, then base classes. there are few edge cases where this matters, not sure if this is one of them

```suggestion
class ConfirmScheduledTimeView(LoginRequiredMixin, LeagueView):
```

heltour/tournament/views.py
```diff
@@ -1801,6 +1801,54 @@ def view(self, post=False):
...
+                                black=player).exclude(scheduled_time=None)]
+
+        context = {
+            'player': player,
+            'active_seasons_with_sp': active_seasons_with_sp,
+            'last_season': last_season,
+            'next_pairings': next_pairings
+        }
+
+        if post:
```
nicwest, line 1834: if this is a class based view, you should be able to overload `get(request)` and `post(request)` independantly.

https://ccbv.co.uk/

glbert-does:
now rebased on the django_update / django 4.2 branch. will get to your comments tomorrow/the day after.

heltour/tournament/lichessapi.py
```diff
@@ -165,6 +165,19 @@ def get_peak_rating(lichess_username, perf_type):
         return None
 
 
+def bulk_start_games(tokens, clock, increment, clockstart, variant, leaguename, priority=0, max_retries=0, timeout=30):
+    # tokens, clock.limit, clock.increment, days, pairAt, startClocksAt, rated, variant, fen, message, rules
+    # tokens: white1:black1,white2:black2,...
+    # clock.limit: in seconds
+    # pairAt/startClocksAt: Unix timestamp in milliseconds
+    # message: has {opponent} and {game} placeholders.
+    # rules: noAbort,noRematch,noGiveTime,noClaimWin,noEarlyDraw
```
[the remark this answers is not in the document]
glbert-does: oh no, i don't think they are important – that's just me copying the lichess api doc so i don't have to switch tabs while writing my code. but now that is done, i don't think it's needed anymore. i think i will rather remove the comment.

nicwest:
Screenshot 2024-12-21 at 13 12 30

heltour/tournament/tasks.py
```diff
@@ -306,6 +312,95 @@ def update_slack_users():
...
+            pass
+
+
+@app.task()
+def start_games():
+    logger.info('[START] Checking for games to start.')
+    games_to_start = PlayerPairing.objects.filter(
+            result='', game_link='',
+            scheduled_time__lt=timezone.now() + timedelta(minutes=5, seconds=30),
+            scheduled_time__gt=timezone.now() + timedelta(seconds=30),
```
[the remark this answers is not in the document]
glbert-does: > can I just check these times are correct?

tl;dr: yes, i did break my pretty little head about those times, and i picked those on purpose. no, i do not think i necessarily picked the best choice.

> would it also not be a good idea to pick up games in the past as well?

i thought about it, but i am honestly not sure. there is a lot of weird stuff that could happen i think. for example, players play a game that is not picked up for some reason and before the mods can add the gamelink, heltour starts a new game for them. that wouldn't be the end of the world of course. but if the game is ongoing, it might distract them.
or maybe players manage to somehow schedule a game for a time that already happened by accident. i think chesster would not allow this, but mods can manually schedule games, and there's less checking there. would still not be the end of the world of course.

at the same time, i do not really see the point in automatically starting games that should have happened a while ago. if the game did not start, or heltour thinks the game did not start, there might be a reason for that, and it might be better to not start it automatically.

anyway, the logic about all time points is definitely debatable. some people might be pissed about games starting 5 minutes early, prohibiting them to play some last minute blitz. same with the question of when exactly to start the clock. currently it's 6 minutes after the game was started, which will be 1 minute after the game was scheduled for most people.

i am absolutely happy to discuss these times, i am not at all confident that i picked the best options. very glad you looked at it.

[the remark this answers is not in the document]
nicwest: That all sounds reasonable, and it can all be adjusted later if needed.

glbert-does:
at this point i think everything is implemented. the only "nice to have" that isn't implemented is logging users out if the token is not valid anymore. but the more i think about that, the less i think it's necessary.

heltour/api_worker/views.py
```diff
@@ -21,10 +21,11 @@ def _get_lichess_api_token():
         return None
 
 def _do_lichess_api_call(redis_key, path, method, post_data, params, priority, max_retries, format,
-                         retry_count=0):
```
lakinwecker, line 24: If this is optional, shouldn't we make it actually optional by setting a default value to `None`?

heltour/tournament/tasks.py
```diff
@@ -311,6 +317,99 @@ def update_slack_users():
...
+        result = lichessapi.bulk_start_games(tokens=tokens, clock=clock, increment=increment, do_clockstart=do_clockstart, clockstart=clockstart, clockstart_in=clockstart_in, variant=variant, leaguename=leaguename)
+    except lichessapi.ApiClientError as err:
+        # try to handle errors due to rjected tokens
+        e = str(err).replace('API failure: CLIENT-ERROR: [400] ', '') # get json part from error
+        try:
+           result = json.loads(e)
+           for bad_token in result["tokens"]:
+               # set expiration of rejected token to yesterday, so we know to not use it anymore.
+               for game in league_games:
+                   if game.white.oauth_token.access_token==bad_token:
```
lakinwecker, line 331: I feel like his inner loop and check should be refactored out to helper methods to make this easier to follow.

heltour/tournament/tasks.py
```diff
@@ -311,6 +317,99 @@ def update_slack_users():
...
+                       game.white.oauth_token.expires = timezone.now() + timedelta(days=-1)
+                       game.white.oauth_token.save()
+                       break # only one oauth_token can be the bad token, so we do not need to proceed the loop
+                   if game.black.oauth_token.access_token==bad_token:
+                       game.black.oauth_token.expires = timezone.now() + timedelta(days=-1)
+                       game.black.oauth_token.save()
+                       break
+               # remove bad token from our token string + the good token paired with it, remove potential superfluous comma
+               new_tokens = re.sub('^,', '', re.sub(f'(^|,)([A-z0-9_]*:)?{bad_token}(:[A-z0-9_]*)?', '', tokens))
```
lakinwecker, line 340: I'm a fan of refactoring out the various patterns in a regex expression and naming them appropriately. I find it makes it easier to figure out what is going on.

heltour/tournament/views.py
```diff
@@ -1813,6 +1815,52 @@ def view(self, post=False):
     def view_post(self):
         return self.view(post=True)
 
+class ConfirmScheduledTimeView(LeagueView, LoginRequiredMixin):
+    def view(self, post=False):
+        player = Player.get_or_create(self.request.user.username)
```
lakinwecker, line 1820: `get_or_create` returns a tuple https://docs.djangoproject.com/en/5.2/ref/models/querysets/#get-or-create

heltour/tournament/views.py
```diff
@@ -1813,6 +1815,52 @@ def view(self, post=False):
...
+
+        active_seasons = self.league.season_set.filter(is_completed=False).order_by('-start_date')
+        active_seasons_with_sp = [(s, player.seasonplayer_set.filter(season=s).first()) for s in
```
lakinwecker, line 1823: There must be a way to do this without individual queries for each season.

heltour/tournament/views.py
```diff
@@ -1813,6 +1815,52 @@ def view(self, post=False):
...
+                                             season__is_active=True)
+        next_pairings = []
+        for r in active_rounds:
+            next_pairings += [(r, p) for p in
+                            r.pairings.filter(white=player).exclude(scheduled_time=None) | r.pairings.filter(
+                                black=player).exclude(scheduled_time=None)]
+
+        context = {
+            'player': player,
+            'active_seasons_with_sp': active_seasons_with_sp,
```
lakinwecker, line 1841: AFAICT, this is not used in the template?

heltour/tournament/views.py
```diff
@@ -1813,6 +1815,52 @@ def view(self, post=False):
...
+            'last_season': last_season,
```
lakinwecker, line 1842: Same with this one

PR #524, opened 2023-11-02
more filters when requesting game metas

### LINKED ISSUE #202: Validate variant in heltour
https://github.com/cyanfish/heltour/blob/master/heltour/tournament/tasks.py#L148

Hasn't been an issue so far but with 960 you want to make sure regular games aren't counted.

### LINKED ISSUE #255: Enhance api use
Use `since` and `vs` parameters to optimize usage

### LINKED ISSUE #498: heltour picks up wrong (old) game
in this pairing S34 R5:
https://www.lichess4545.com/admin/tournament/playerpairing/77380/change/
heltour accidentally picked up the wrong game – https://lichess.org/LQJheBZu – which was actually played months earlier for S33 R6:
https://www.lichess4545.com/admin/tournament/playerpairing/75410/change/
(same players, same colours)

this is a rare bug, definitely does not happen every time we get a repeat pairing. i believe i remember it happening before once, but i might be mistaken.

we fixed the problem manually by replacing the game link.

glbert-does:
closes #498. hopefully.

for games that are supposed to start soon, we requested the last games of the white player from lichess. we now filter these to only request games since the round started + only against the black player.

while i am still not certain how the bug above worked exactly, i assume that this should fix it. otherwise, i am out of ideas.

edit: rebased on #527

more edits: i believe this closes #255 as well.

edit to the edits: closes #202 too.

### DIFF
--- a/heltour/tournament/lichessapi.py
+++ b/heltour/tournament/lichessapi.py
@@ -111,9 +111,9 @@ def get_game_meta(gameid, priority=0, max_retries=5, timeout=1800):
     return json.loads(result)
 
 
-def get_latest_game_metas(*, lichess_username, since, number, opponent, priority=0, max_retries=5, timeout=1800):
-    url = f'{settings.API_WORKER_HOST}/lichessapi/api/games/user/{lichess_username}?since={since}&max={number}\
-            &vs={opponent}&ongoing=true&priority={priority}&max_retries={max_retries}&format=application/x-ndjson'
+def get_latest_game_metas(*, lichess_username, since, number, opponent, variant, priority=0, max_retries=5, timeout=1800):
+    url = (f'{settings.API_WORKER_HOST}/lichessapi/api/games/user/{lichess_username}?since={since}&max={number}'
+           f'&vs={opponent}&perfType="{variant}"&ongoing=true&priority={priority}&max_retries={max_retries}&format=application/x-ndjson')
     result = _apicall_with_error_parsing(url, timeout)
     return [json.loads(g) for g in result.split('\n') if g.strip()]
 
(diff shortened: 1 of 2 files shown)


heltour/tournament/lichessapi.py
```diff
@@ -111,9 +111,9 @@ def get_game_meta(gameid, priority=0, max_retries=5, timeout=1800):
     return json.loads(result)
 
 
-def get_latest_game_metas(lichess_username, number, priority=0, max_retries=5, timeout=1800):
-    url = '%s/lichessapi/api/games/user/%s?max=%s&ongoing=true&priority=%s&max_retries=%s&format=application/x-ndjson' % (
-        settings.API_WORKER_HOST, lichess_username, number, priority, max_retries)
+def get_latest_game_metas(lichess_username, since, number, opponent, priority=0, max_retries=5, timeout=1800):
```
nicwest, line 114:
```suggestion
def get_latest_game_metas(*, lichess_username, since, number, opponent, priority=0, max_retries=5, timeout=1800):
```

positional arguments are the devil, use `*` to enforce keyword arguments and make your code more readable/less error prone!

heltour/tournament/lichessapi.py
```diff
@@ -111,9 +111,9 @@ def get_game_meta(gameid, priority=0, max_retries=5, timeout=1800):
...
+    url = '%s/lichessapi/api/games/user/%s?since=%s&max=%s&vs=%s&ongoing=true&priority=%s&max_retries=%s&format=application/x-ndjson' % (
+        settings.API_WORKER_HOST, lichess_username, since, number, opponent, priority, max_retries)
```
nicwest, line 116: use `.format()` or preferable a format string `f"{settings.API_WORKER_HOST}/lichessapi..."`

nicwest: https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals

nicwest (commented):
LGTM, some minor cosmetic things

glbert-does:
implemented nic's comments and rebased from branch django_update

PR #527, opened 2023-11-23
update django to 4.2

### LINKED ISSUE #525: pyOpenSSL 22.0.0 is incompatible with cryptography 41.0.4
current master is broken, because cryptography 41.0.4 requires pyOpenSSL > 39.

glbert-does:
* if you do not want cacheops in debug mode, you should set `CACHEOPS_ENABLED = False` in your settings now.
* django-select2-forms is not compatible with django 4+. the way i understand it, it has not been developed for a while, since django 2 supposedly supports select2 directly. ~~i don't really see where we even use select2, so i simply removed it.~~ removing it is not possible, unless we squash migrations.

edit: this will also close #525


glbert-does:
just noticed that django 5 was released in the meantime. will soon work on upgrading.

heltour/tournament/admin.py
```diff
@@ -544,7 +543,7 @@ def review_nominated_games_pgn_view(self, request, object_id):
         pgn = lichessapi.get_pgn_with_cache(gameid, priority=10)
 
         # Strip most tags for "blind" review
-        pgn = re.sub('\[[^R]\w+ ".*"\]\n', '', pgn)
+        pgn = re.sub('\\[[^R]\\w+ ".*"\\]\n', '', pgn)
```
nicwest, line 546: but I think the pythonic way of doing this is using a regrex string literal:
```suggestion
        pgn = re.sub(r'\[[^R]\w+ ".*"\]\n', '', pgn)
```
> The solution is to use Python’s raw string notation for regular expression patterns; backslashes are not handled in any special way in a string literal prefixed with 'r'. So r"\n" is a two-character string containing '\' and 'n', while "\n" is a one-character string containing a newline. Usually patterns will be expressed in Python code using this raw string notation.

https://docs.python.org/3/library/re.html

heltour/tournament/models.py
```diff
@@ -732,7 +731,7 @@ def __str__(self):
         return self.account_username
 
 
-username_validator = RegexValidator('^[\w-]+$')
+username_validator = RegexValidator('^[\\w-]+$')
```
nicwest, line 734: see above on regexp litrals

heltour/tournament/signals.py
```diff
@@ -1,53 +1,50 @@
 from django.dispatch.dispatcher import Signal
 
 # Signals that run tasks
-do_generate_pairings = Signal(providing_args=['round_id', 'overwrite'])
-do_round_transition = Signal(providing_args=['round_id'])
-do_schedule_publish = Signal(providing_args=['round_id', 'eta'])
-do_pairings_published = Signal(providing_args=['round_id'])
-do_validate_registration = Signal(providing_args=['reg_id'])
-do_create_team_channel = Signal(providing_args=['team_ids'])
+do_generate_pairings = Signal() # providing_args=['round_id', 'overwrite']
```
nicwest, line 4: delete comments, we have version control.

heltour/tournament/tests/test_login.py
```diff
@@ -3,7 +3,8 @@
 from django.test import TestCase
 from unittest.mock import patch
 from heltour.tournament import oauth
-from .testutils import *
+from .testutils import *#
```
nicwest, line 6: generally speaking it's worth  being explicit about your imports. from a practical perspective it helps static analysis because when you have `import *` the program (it it's not like a LSP or something) really struggles to know what variables/functions etc are on the stack.

Additionally relative imports are the devil, prefere absolute imports if possible.

heltour/tournament/tests/test_login.py
```diff
@@ -20,13 +21,18 @@ def test_encode_decode_state(self, *args):
     @patch('heltour.tournament.oauth._encode_state', return_value='encodedstate')
     def test_oauth_redirect(self, *args):
         response = self.client.get(league_url('team', 'login'))
-        expected_oauth_url = 'https://oauth.lichess.org/oauth/authorize' + \
+        url = re.sub("&code_challenge=[0-9A-z-]{43}", "", response.url)
+        expected_oauth_url = 'https://lichess.org/oauth' + \
```
nicwest, line 25:
```suggestion
        expected_oauth_url = ('https://lichess.org/oauth'
                             '?response_type=code'
                             '?response_type=code'
                             '&client_id=clientid'
                             '&client_id=heltour'
                             '&redirect_uri=http://testserver/auth/lichess/'
                             '&redirect_uri=https://testserver/auth/lichess/'
                             '&scope=email:read%20challenge:read%20challenge:write'
                             '&scope=email:read%20challenge:read%20challenge:write'
                             '&code_challenge_method=S256'
                             '&state=encodedstate'
                             '&state=encodedstate')
```
python will concat this strings for you.

I would also make an argument to use https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse as you are looking at a URL. but not the end of the world

nicwest (approved):
LGTM, a few minor cosmetic things @lakinwecker can we get this merged?

glbert-does:
> LGTM, a few minor cosmetic things @lakinwecker can we get this merged?

i'll fix the minor stuff on tuesday/wednesday.

lakinwecker:
I trust you all more than I trust myself on this stuff at the moment. I won't have time to review it in depth, anytime soon. Do you need me to mash the merge button? Or does one of you have the perms to do that? 

I can deploy whenever you are ready

nicwest:
I ran into an error testing this locally so don't merge yet.

Django 4.1 upgraded the `ManifestStaticFilesStorage` method for collecting static files and that's playing havok with our mapless bootstrap files (I suspect that's the reason that the tests are using the more tolerant `StaticFilesStorage`).

ref: https://github.com/glbert-does/heltour/blob/70a1edcbec3923029412ad179967cc6e5e49097a/heltour/tournament/static/lib/css/bootstrap.min.css#L6

glbert-does:
i wonder why i did not have that local error.
do i understand this correctly that you're saying a line like the one you pointed to needs to be added to all css files that are missing them?
so e.g. `heltour/tournament/static/lib/css/BootstrapXL.css`, which does not have it currently
needs `/*# sourceMappingURL=BootstrapXL.css.map */`?

i will fix this tomorrow/day after tomorrow as well. also, i will try to see why i didn't have the error locally.

PS: lakin, you gave me the merge perm. but the only way for me to use it responsibly is to not use it.

nicwest:
> do i understand this correctly that you're saying a line like the one you pointed to needs to be added to all css files that are missing them?

no. I'm suggesting that some of the current files are missing maps for their minified content but are still referencing them. The old version of `ManifestStaticFilesStorage`  was permissive of this, but the new version isn't.

I'm not sure what settings you are running locally but if you change the staticfiles finder to `ManifestStaticFilesStorage` and try `python manage.py collectstatic` that should produce the error I was seeing. off the top of my head it was complaining about no being able to find `/lib/css/bootstrap.css.min.map` or similar

glbert-does:
@nicwest fixed all the minor issues you commented about above.

i also tried to read up on the `ManifestStaticFilesStorage`. and i did create the map files. i understand more about static file storage now (easy, as opposed to zero before), but i won't pretend to fully comprehend why empty files would be needed.

nicwest:
you can grab the files from here:  https://stackpath.bootstrapcdn.com/bootstrap/<version>/css/<name of file>
https://github.com/glbert-does/heltour/blob/2d83b28015ea49240f94d481f56860c75077b6a8/heltour/tournament/static/lib/css/bootstrap-theme.min.css#L2C15-L2C20 the version is at the top of the file.

glbert-does:
right. i added those map files, but this is some dark magic. locally, the empty files did not lead to an error for me already, so i am not convinced i fixed it.

i turned debug on and off again, and switched around some folders and STATIC*-settings, and in the end i did manage to break everything. doesn't seem related to those map files though, so i am still in the dark.

nicwest:
you may need to to run `poetry lock --no-update`

PR #533, opened 2024-03-29
slightly change sorting in teamgen

glbert-does:
this change in sorting only makes a difference for players who picked both alt_fine, and were previous_season_alts; and there only sometimes. but when it does make a difference, we probably prefer this changed sorting.


nicwest (approved):
LGTM

PR #537, opened 2024-06-17
Move to using docker for dependencies and deployments

nicwest:


nicwest:
summary of stuff that changes.

- settings rationalised into base, development and production
- docker files
  - heltour:latest: contains the python env (the webapp, celery, and eventually the api worker)
  - heltour-static: a static files server. at somepoint we should talk about moving this stuff to a CDN.
- docker compose to replace vagrant. I've been using this locally and with the new docker compose `watch` feature, this feels viable.
- includes glbert's #527 
  
still todo:
- api worker
- secrets
- figuring out what to do with database stuff. I'm playing around with django fixtures and feels viable.

Issue #358, opened 2020-10-10
Integrate Prospector


lakinwecker:
We should have and use a code linting tool, like Prospector.

PR #543, opened 2024-12-15
Ruff linting/formatting

nicwest:
https://github.com/astral-sh/ruff

Ruff is an opinionated (but configurable) linter/fixer.

I'm going to make an argument that we should have a consistent and enforceable style for this project. this will hopefully save a lot of review hours going forward. (it should also help us find bugs and smelly code)

this PR includes #527  and #537  and probably should only be merged after those two.


glbert-does:
this would presumably close #358.

lakinwecker:
Closing this one. We're moving forward with this suggestion (thanks @nicwest ) but we are doing it in a different PR

PR #544, opened 2024-12-22
implement zen mode

### LINKED ISSUE #541: hide all ratings on the site
there's always people who get stressed out by ratings, so lichess itself has an option to hide all ratings from the site here: https://lichess.org/account/preferences/display

it would be nice if lichess4545 had such an option too.

glbert-does:
closes #541

it implements zen mode using template tags. i think there are likely good arguments to do it using css instead, but i don't think the code would have been much simpler, since there is currently nothing that marks something as a rating consistently, so one would still have to add a lot of html attributes. possibly using css would be more future proof, but then again, there is nothing preventing anyone not using a new html attribute for newly added ratings.
that is why i decided to hide ratings completely using tags.

i still think the main work here is to find all places that reference ratings. this should be mostly done, but of course it is plausible that i missed some places.

i decided to not remove ratings from the stats pages. they are almost completely anonymous there. but maybe one should? i prefer to wait for zen mode users to complain about it.


lakinwecker:
This actually looks pretty good to me in terms of a feature.  I think it's interesting though, because my instincts want to try and find a better way to implement it, but I'm also unsure if there really is

glbert-does:
my instincts are the same, which is why i argued against the css-strawman in the initial pr comment. it feels like there has to be a better way, though i don't know what it is.

PR #546, opened 2024-12-27
only approved users need to link

### LINKED ISSUE #268: Only show account linking message for invited users

### LINKED ISSUE #538: dashboard for players who are not approved asks them to link
on the dashboard any unlinked player is asked to link their account:
https://github.com/Lichess4545/heltour/blob/master/heltour/tournament/templates/tournament/user_dashboard.html#L9-L19

unlinked players includes those who were not invited to slack, which is confusing for them. ideally, i think we'd add another condition to `{% if not slack_linked %}` – for example we could check for the existence of any season player object.
short term, maybe the text should just qualify with something like "to participate in the league, after being invited to our slack workspace, you need to link ..."

glbert-does:
only users who have at least one approved registration are asked to link their account on the user dashboard.

definitely closes #538, also closes #268 if i understand that one correctly.

based on #527.

### DIFF
--- a/heltour/tournament/templates/tournament/user_dashboard.html
+++ b/heltour/tournament/templates/tournament/user_dashboard.html
@@ -6,7 +6,7 @@
 {% block content %}
     <div class="row row-condensed-xs home-row">
         <div class="col-md-6 col-xs-12">
-            {% if not slack_linked %}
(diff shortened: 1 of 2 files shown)


lakinwecker (approved):
LGTM

PR #551, opened 2025-01-10
remove captcha

glbert-does:
this will remove the captchas on the registration and the contact page, and remove django_recaptcha from the dependencies.
i believe captcha is not necessary anymore, since we require oauth login to use either of these pages.

based on #527.


lakinwecker:
Closed in favour of mine, where I updated the poetry.lock

PR #555, opened 2025-01-29
only change registration status to pending when some specific fields are edited

### LINKED ISSUE #554: do not automatically set registration to pending after edits
the way i see it, of all the fields in the registration form, mods should only have to review a registration again if any of:
* alternate_preference
* weeks_unavailable
* agreed_to_rules
* agreed_to_tos
* section_preference

changed. then setting the status to pending makes sense.
with most other fields mods shouldn't have to review changes – specifically:
* friends
* avoid

don't need mod review, so if those change registration status should not be set to "pending" again.

https://github.com/Lichess4545/heltour/blob/15dce1b2ab6bc21822ed258472396161933c4d41/heltour/tournament/forms.py#L212

glbert-does:
only change registration status to pending when some specific fields are changed, that require mod review.

closes #554. as opposed to 554, we do not change status to pending when agreed_to_rules
or agreed_to_tos are changed, because, and this is true, they cannot be changed anyway.

### DIFF
--- a/heltour/tournament/forms.py
+++ b/heltour/tournament/forms.py
@@ -203,7 +203,8 @@ class RegistrationForm(forms.ModelForm):
         registration = super(RegistrationForm, self).save(commit=False, *args, **kwargs)
         registration.season = self.season
         registration.lichess_username = str(self.username)
-        registration.status = 'pending'
+        if any(x in self.changed_data for x in ['alternate_preference', 'section_preference', 'weeks_unavailable']):
+            registration.status = 'pending'
         if commit:
             registration.save()
         registration.player().agreed_to_tos()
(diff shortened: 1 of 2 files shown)


heltour/tournament/forms.py
```diff
@@ -203,7 +203,8 @@ def save(self, commit=True, *args, **kwargs):
         registration = super(RegistrationForm, self).save(commit=False, *args, **kwargs)
         registration.season = self.season
         registration.lichess_username = str(self.username)
-        registration.status = 'pending'
+        if any(x in self.changed_data for x in ['alternate_preference', 'section_preference', 'weeks_unavailable']):
```
lakinwecker, line 206: Minor nit, use a set:
`if set(self.changed_data) & {'alternate_preference', 'section_preference', 'weeks_unavailable'}:`

lakinwecker (changes requested):
Minor nit for readability.

glbert-does:
requested changes are in pr #610

PR #557, opened 2025-02-04
fix naive datetime warning

glbert-does:
fix an annoying warning about player.date_modified receiving a naive datetime

based on the updated django branch #527


lakinwecker (approved):
LGTM

PR #583, opened 2025-03-17
split big registration task into many small ones

### LINKED ISSUE #582: split up task of validating registrations
the daily task of validating registrations makes us hit lichess api limits, which in turn could be a problem for celery.

so instead of running one huge daily task for validating registrations: every n minutes take the m registrations that were checked the longest time ago and validate them. n=30 minutes and m=5 registrations for example would validate 240 registrations per day. should probably make m an option in the settings. n is an option in the settings anyway.

glbert-does:
closes #582 by splitting the single huge registration validation task into many small ones, and validating one single reg every 5 minutes.


heltour/tournament/migrations/0002_registration_last_validation_try.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('tournament', '0001_squashed_0195'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='registration',
+            name='last_validation_try',
+            field=models.DateTimeField(default=django.utils.timezone.now),
```
lakinwecker, line 17: Does this mean they all start with `now` as the value when we start?

glbert-does: yes, they will start with `now`. i guess we could pick any other value too, as long as it is not in the future.

heltour/tournament/tasks.py
```diff
@@ -502,10 +504,9 @@ def do_validate_registration(reg_id, **kwargs):
 @app.task()
 def validate_pending_registrations():
     # we want to re-validate pending registrations if they are not marked valid already
-    regs_to_validate = Registration.objects.filter(season__registration_open=True, status__exact='pending').exclude(Q(validation_warning=False) & Q(validation_ok=True))
-    logger.info(f'Validating {len(regs_to_validate)} player registrations')
-    for reg in regs_to_validate:
-        signals.do_validate_registration.send(sender=validate_pending_registrations, reg_id=reg.pk)
+    reg_to_validate = Registration.objects.filter(season__registration_open=True, status__exact='pending').exclude(Q(validation_warning=False) & Q(validation_ok=True)).order_by('last_validation_try').first()
```
lakinwecker, line 507: Why are we excluding validation_warning=False ?

heltour/tournament/migrations/0002_registration_last_validation_try.py
lakinwecker: Yeah, I think it's fine?  Will this queue  ever be "complete"? or will there always be more to process?

heltour/tournament/tasks.py
lakinwecker: Yeah, ok the validation_warning makes sense.

heltour/tournament/migrations/0002_registration_last_validation_try.py
glbert-does: never complete, every 5 minutes it will try to validate a registration, if there is at least one that was not successfully validated yet (in any season that has open registrations). that is why i was considering filtering for a minimum time delta too, so that re-validation of the same reg is only attempted after, say, 12 hours.
in the extreme case of only one not successfully validated reg, we would currently re-validate that one every 5 minutes.
the only way it would probably ever be complete would be if there was not season with open registrations. which might happen now and then, but not all too often.

lakinwecker: Yeah, I like filtering for "hasn't been updated in 12 hours".

lakinwecker (changes requested):
Let's filter for a minimum time since we last tried to validate it

glbert-does:
done. i did pick 24 hours though instead of 12, that seems easily enough.

heltour/tournament/tasks.py
```diff
@@ -438,6 +438,8 @@ def do_generate_pairings(sender, round_id, overwrite=False, **kwargs):
 @app.task()
 def validate_registration(reg_id):
     reg = Registration.objects.get(pk=reg_id)
+    reg.last_validation_try = timezone.now()
```
lakinwecker, line 441: Probably best to rewrite this as an update, rather than a `.save()`  Something like:
`Registration.objects.filter(pk=reg_id).update(last_validation_try=timezone.now()`

This means that ONLY the last_validation_try will be updated at this point, which is slightly safer. Practically it probably doesn't make any difference, but it's a good habit to have.

PR #589, opened 2025-03-25
change season ordering

glbert-does:
change the ordering of seasons in the tournament admin to show seasons that are not completed first, since those will usually be the more interesting seasons.


lakinwecker (approved):
LGTM

PR #591, opened 2025-03-27
fix debug-toolbar on vagrant

glbert-does:
vagrant uses 10.0.2.2 as the ip for the host system on the private network, so when accessing heltour on the guest from a browser on the host, this ip needs to be considered internal


lakinwecker (approved):
LGTM

PR #594, opened 2025-04-02
remove revision, use update, add additional except

glbert-does:
* remove revision for reg validation
* catch an additional api worker error
* with revision removed, we do not need to use save() anymore, use update() instead


lakinwecker (approved):
LGTM

PR #602, opened 2025-04-13
fix problem with zero round_list length

### LINKED ISSUE #601: trying to edit availability after season has ended leads to server error
in views.py

glbert-does:
fixes #601


lakinwecker (approved):
LGTM

PR #605, opened 2025-04-14
catch oauth badsignature

glbert-does:
avoiding some internal server errors


lakinwecker (approved):
LGTM. Probably bots, or hackerone script kiddies

PR #623, opened 2025-05-06
link autocreated games in a slack game channel

glbert-does:
fixing the problem where chesster does not mention games that were autocreated in the game channel.
it's plausible that this would be better to fix on the chesster side.

(this is based on #622, which is more urgent.)


lakinwecker (approved):
I'd honestly prefer to have more of this in heltour and less in chesster IMO

heltour/tournament/migrations/0005_league_channels.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 4.2.20 on 2025-05-06 19:59
+
```
lakinwecker, line 2: You'll need a merge migration for this now.

lakinwecker (changes requested):
needs a merge migration now

lakinwecker (approved):
LGTM

PR #631, opened 2025-05-09
fix some tests, add some tests, disable some tests

glbert-does:
work towards more tests. starting with tests for code i wrote, then adding some low-hanging fruits where tests are easy to write.

edit: note that all the tests disabled here did ***not*** work before this pr anyway.

edit 2: note that `python manage.py test heltour.tournament.tests.test_models.LonePlayerPairingTestCase.test_scheduling_creates_notification` does not work without a fix for notifications, like e.g. #622.

edit 3: also fixes a bug in get_access_token that i noticed while writing the tests.

edit 4: also note that one of the included tests does ***not*** pass without #632 being applied first.


glbert-does:
now also fixes a but in get_access_token that i noticed while writing the tests.

heltour/tournament/models.py
```diff
@@ -904,7 +904,10 @@ def agreed_to_tos(self):
         )
 
     def get_access_token(self):
-        return self.oauth_token.access_token
+        if self.oauth_token is not None:
+            return self.oauth_token.access_token
+        else:
+            return None
```
lakinwecker, line 910: nit: I slightly prefer this style of guard

```suggestion
        if self.oauth_token is None:
            return None
        return self.oauth_token.access_token
```

heltour/tournament/tests/testutils.py
```diff
@@ -1,5 +1,5 @@
-from heltour.tournament.models import (League, LonePlayerScore, Player, Registration, Season,
```
lakinwecker, line 1: `Round` doesn't appear to be used in this file.

heltour/tournament/tests/test_views.py
```diff
@@ -93,7 +125,12 @@ def setUp(self):
         createCommonLeagueData()
 
     def test_template(self):
-        response = self.client.get(season_url('team', 'wallchart'))
+        # triggering a 404 writes to the log, disable that temporarily for nicer test output
```
lakinwecker, line 128: You don't have to fix this, but we should consider a nicer logging setup in general here.

heltour/tournament/tests/test_views.py
```diff
@@ -84,7 +111,12 @@ def test_template(self):
         response = self.client.get(season_url('team', 'crosstable'))
         self.assertTemplateUsed(response, 'tournament/team_crosstable.html')
 
-        response = self.client.get(season_url('lone', 'crosstable'))
+        # triggering a 404 writes to the log, disable that temporarily for nicer test output
+        logging.disable(logging.CRITICAL)
```
lakinwecker, line 115: Ok, yeah, with the heavy use, you should consider a context manager to disable logs so you can do:
```
with noLogging():
    doSomething()
```

lakinwecker (approved):
Approving with some minor comments, will merge to go live today

heltour/tournament/tests/testutils.py
glbert-does: it's used in `get_rounds` - which itself is not used anywhere, yet, but i am already using it in a local test that is not in this pr yet.

PR #632, opened 2025-05-10
code refactoring in autostart games

### LINKED ISSUE #627: code refactoring in autostart games
in #523, lakin requested refactoring of some code.

glbert-does:
closes #627.
based on #623, which is based on #622.

### DIFF
--- a/heltour/tournament/models.py
+++ b/heltour/tournament/models.py
@@ -890,10 +890,11 @@ class Player(_BaseModel):
             date_first_agreed_to_tos=now
         )
 
-
     def get_access_token(self):
         return self.oauth_token.access_token
 
+    def token_valid(self):
+        return self.oauth_token.access_token not None and not self.oauth_token.is_expired()
 
     def __str__(self):
         if self.rating is None:
@@ -1511,6 +1512,9 @@ class PlayerPairing(_BaseModel):
 
     def get_black_access_token(self):
         return self.black.get_access_token()
+
+    def tokens_valid(self):
+        return self.white.token_valid() and self.black.token_valid()
 
     def black_display(self):
         if not self.black:
--- a/heltour/tournament/models.py
+++ b/heltour/tournament/models.py
@@ -900,7 +900,7 @@ class Player(_BaseModel):
         return self.oauth_token.access_token
 
     def token_valid(self):
-        return self.oauth_token.access_token not None and not self.oauth_token.is_expired()
+        return self.oauth_token.access_token is not None and not self.oauth_token.is_expired()
 
     def __str__(self):
         if self.rating is None:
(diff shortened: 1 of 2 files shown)


heltour/tournament/tasks.py
```diff
@@ -317,6 +320,19 @@ def update_slack_users():
...
 
 
+def _expire_bad_tokens(*, league_games, bad_token):
+    # set expiration of rejected token to yesterday, so we know to not use it anymore.
+    for game in league_games:
+        if game.get_white_access_token()==bad_token:
+            game.white.oauth_token.expires = timezone.now() + timedelta(days=-1)
+            game.white.oauth_token.save()
+            return None # only one oauth_token can be the bad token, so we do not need to proceed the function
+        if game.get_black_access_token==bad_token:
```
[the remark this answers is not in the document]
glbert-does: thanks and done

heltour/tournament/models.py
```diff
@@ -890,10 +896,11 @@ def agreed_to_tos(self):
             date_first_agreed_to_tos=now
         )
 
-
     def get_access_token(self):
         return self.oauth_token.access_token
 
+    def token_valid(self):
+        return self.oauth_token.access_token is not None and not self.oauth_token.is_expired()
```
lakinwecker, line 903: Use the `get_access_token` function above, or, better yet, add a function to `oauth_token` which does this full check

heltour/tournament/tasks.py
```diff
@@ -317,31 +320,41 @@ def update_slack_users():
...
+    # set expiration of rejected token to yesterday, so we know to not use it anymore.
+    for game in league_games:
+        if game.get_white_access_token()==bad_token:
+            game.white.oauth_token.expires = timezone.now() + timedelta(days=-1)
+            game.white.oauth_token.save()
+            return None # only one oauth_token can be the bad token, so we do not need to proceed the function
+        if game.get_black_access_token()==bad_token:
+            game.black.oauth_token.expires = timezone.now() + timedelta(days=-1)
+            game.black.oauth_token.save()
+            return None
```
lakinwecker, line 333: I want to do something like the below (which would require some more or different getter methods):

```suggestion
        for token in [game.get_white_oauth_token(), game.get_black_oauth_token()]:
        if oauth_token.access_token==bad_token:.
            oauth_token.expires = timezone.now() + timedelta(days=-1)
            oauth_token.save()
            break # only one oauth_token can be the bad token, so we do not need to proceed the function
```

lakinwecker (changes requested):
These are nitpicky, but I figured they were worth it from a "refactoring" perspective.

lakinwecker (approved):

PR #635, opened 2025-05-13
fix internal server error on admin search

### LINKED ISSUE #633: internal server error when using the admin search field for schedulednotification
Cannot resolve keyword 'player' into field.

glbert-does:
and add a test. closes #633.


lakinwecker (approved):
LGTM

PR #646, opened 2025-05-20
fix some problems appearing with the shortened registration

glbert-does:
1. fix a bug where reg validations still expected there to be the "are you on slack" question on the form.
2. add test cases for validation
3. fix a small bug with missing brackets in the notifications to mods
4. fix a small bug where ApiWorkerError and ApiClientError were caught in such an order, that ApiClientErrors were always caught by the ApiWorkerError already
5. add and use a context manager for disabling logging in tests
6. fix tests that added registrations without an existing player (not possible anymore, because registering requires login, and login adds a player if it does not exist)

### DIFF
--- a/heltour/tournament/tests/testutils.py
+++ b/heltour/tournament/tests/testutils.py
@@ -1,3 +1,4 @@
+import logging
 from heltour.tournament.models import (League, LonePlayerScore, Player, Registration, Round,
                                        Season, SeasonPlayer, Team, TeamMember, TeamScore)
 from django.urls import reverse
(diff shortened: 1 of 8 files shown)


heltour/tournament/tests/testutils.py
```diff
@@ -74,3 +75,17 @@ def createCommonLeagueData(round_count=3):
             LonePlayerScore.objects.create(season_player=sp)
             player_num += 1
             TeamMember.objects.create(team=team, player=player, board_number=b)
+
+
+class Shush:
+    def __init__(self):
+        logging.disable(logging.CRITICAL)
```
lakinwecker, line 82: Probably this should be in the enter method, not the init method

lakinwecker (approved):

PR #650, opened 2025-05-21
more tests from the testing branch

glbert-does:
in this pr included:
* fix context manager Shush() so logging is turned off in __enter__
* in many cases use setUpTestData instead of setUp, to speed up tests
* in some cases, combine tests into one class where it makes sense (e.g. Template testing)
* a test for #649
* generally more tests


lakinwecker (approved):
LGTM

PR #652, opened 2025-05-23
fix EditRostersForm

### LINKED ISSUE #651: manage_players does not allow moving players to alternate
or rather, moving them does not persist on save. possibly caused by the registration changes? i will investigate.

glbert-does:
fix #651, add a test.

remove expected ratings in some other places. i hope the changes in edit_rosters.js are correct, i don't know all that much about those js files. locally, it seems to work.

### DIFF
--- a/heltour/tournament/static/tournament/js/edit_rosters.js
+++ b/heltour/tournament/static/tournament/js/edit_rosters.js
@@ -134,18 +134,11 @@ function updateAverage($teams) {
         var $players = $team.find('.player');
         var n = 0;
         var total = 0.0;
-        var nExp = 0;
-        var totalExp = 0.0;
         for (var j = 0; j < $players.length; j++) {
             var rating = parseInt($players.eq(j).attr('data-rating'));
             if (!isNaN(rating)) {
                 n += 1;
                 total += rating;
-            }
-            var ratingExp = parseInt($players.eq(j).attr('data-exp-rating'));
-            if (!isNaN(ratingExp)) {
-                nExp += 1;
-                totalExp += ratingExp;
             }
         }
         if (n > 0) {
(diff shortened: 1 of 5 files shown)


heltour/tournament/static/tournament/js/edit_rosters.js
```diff
@@ -134,35 +134,23 @@ function updateAverage($teams) {
...
-            $team.find('.average-exp-rating').text((totalExp / nExp).toFixed(2));
-        } else {
-            $team.find('.average-exp-rating').text('');
-        }
     }
 }
 
 function updateDisplayedRatings() {
-    attrName = $('#id_rating_type').val() === 'expected' ? 'data-exp-rating' : 'data-rating';
+    attrName = $('#id_rating_type').val() === 'data-rating';
```
lakinwecker, line 153: This change is not correct. The previous one was checking to see if that rating type form element was set to 'expected'. If it was, it would set the `attrName` variable equal to `'data-exp-rating'. If not, it would set it to `'data-rating`

The new version sets attrName to `false`. becuase `id_rating_type` doesn't exist and  the value will be undefined which does not strictly equal 'data-rating'.

heltour/tournament/templates/tournament/admin/edit_rosters.html
```diff
@@ -55,7 +55,6 @@ <h2>Teams</h2>
                     <th>Board {{ board_number }}</th>
                 {% endfor %}
                 <th>Average</th>
-                <th>Expected</th>
```
lakinwecker, line 58: If you remove this header, don't we have to remove columns below as well?

lakinwecker (changes requested):
The javascript is definitely wrong, although I'm not sure how serious that is.

the template just feels wrong, although I did not verify it.

PR #653, opened 2025-05-25
add missing suspiciousoperation import, catch noreversematch

glbert-does:
also catch NoReverseMatch. and add a test.


lakinwecker (approved):
LGTM

PR #654, opened 2025-05-26
fix the edit_rosters css

glbert-does:
also remove the empty columns for alternates and unassigned players; and fix the css so only the last column is more narrow, not the last two.


lakinwecker (approved):
LGTM

PR #655, opened 2025-05-26
don't add doc_url to context of registerview when post

glbert-does:
fixes an internal server error for invalid registration forms.
add a test for it.


lakinwecker (changes requested):
I think this scoping setup is much too subtle, let's make it less likely to error by explicitly setting the doc_url at the top of the function

lakinwecker (approved):

PR #656, opened 2025-05-28
use get_or_create in User-creation in oauth

glbert-does:
hopefully improving the situation with a race condition.

also add a test to make sure user passwords are unusable.

### DIFF
--- a/heltour/tournament/oauth.py
+++ b/heltour/tournament/oauth.py
@@ -68,7 +68,7 @@ def login_with_code(request, code, encoded_state):
     player.save()
 
     # a password starting with ! signifies an unusable password to django
-    user, = User.objects.get_or_create(username=username, defaults={"password": f"!{create_api_token(length=40)}"})
+    user, _ = User.objects.get_or_create(username=username, defaults={"password": f"!{create_api_token(length=40)}"})
     user.backend = 'django.contrib.auth.backends.ModelBackend'
     login(request, user)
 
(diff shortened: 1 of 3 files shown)


heltour/tournament/oauth.py
```diff
@@ -59,23 +60,19 @@ def login_with_code(request, code, encoded_state):
...
     player.save()
 
-    user = User.objects.filter(username__iexact=username).first()
-    if not user:
-        # Create the user with a password no one will ever use; it can always be manually reset if needed
-        with reversion.create_revision():
-            reversion.set_comment('Create user from lichess OAuth2 login')
-            user = User.objects.create_user(username=username, password=create_api_token())
+    # a password starting with ! signifies an unusable password to django
+    user, _ = User.objects.get_or_create(username=normalize('NFKC', username), defaults={"password": f"!{create_api_token(length=40)}"})
```
lakinwecker, line 69: Make the password generation be a function with some notes/documentation (like the comment above).  Same with the username generation. Making them be their own function and putting all of the context discussed into the comments will help future us understand these decisions.

The functions aren't strictly necessary, but they help contextualize the comments. The functions can even be defined inline here, just makes it easier to see that "These comments are about the password generation" and "these comments are about username generation", both of which are orthogonal concerns to the `User.objects.get_or_create(`

lakinwecker (changes requested):
This generally looks good, but I think that the comments / separate functions for username/password will help us in the long run. These functions can also be testable if we ever need to in the future.

lakinwecker (approved):
LGTM

PR #657, opened 2025-05-29
improvements for the active player table

glbert-does:
* directly use a SQL query to get the count of pairings per player
* in the same query, get the last time they played (or rather scheduled a pairing that has a link, but that should be almost the same thing) – instead of the last season
* put a summary of total games played and total players on the first page.


heltour/tournament/views.py
```diff
@@ -1234,56 +1234,42 @@ def _view(league_tag, season_tag, user_data):
...
 
         paginator = Paginator(tablesums, DEFAULT_PAGE_SIZE)
         page_obj = paginator.get_page(page)
 
+        pks = [player.player_id for player in page_obj.object_list]
+        seasondatafull = SeasonPlayer.objects.filter(player__pk__in = pks, season__league=self.league).values("player__pk").annotate(
+                season_count = Count("player__pk"),
+                ).values("player__lichess_username", "player__pk", "season_count")
+
+        seasondata = {f'{pl["player__pk"]}': [pl["player__lichess_username"], pl["season_count"]] for pl in seasondatafull}
```
lakinwecker, line 1249: why convert to a string here?

lakinwecker: the key of the dict can be an int

heltour/tournament/views.py
```diff
@@ -1234,56 +1234,42 @@ def _view(league_tag, season_tag, user_data):
...
+
+        oneplayer = namedtuple('oneplayer', ['game_count', 'lichess_username', 'season_count', 'last_played'])
+        subtable = []
+
+        for player in page_obj.object_list:
+            subtable.append(oneplayer._make([player.game_count, *seasondata[str(player.player_id)], player.last_played]))
```
lakinwecker, line 1255: avoids converting to a string here to

lakinwecker (changes requested):
One nit

heltour/tournament/views.py
glbert-does: good questions. i think i initially had the username as the key, and then instead of removing the string conversion here, added another conversion below.

lakinwecker (approved):
LGTM

PR #658, opened 2025-05-31
mostly tests for admin.py from the testing branch

glbert-does:
add a couple more tests.

this one also contains pr #655.


glbert-does:
this pr got too big, i will split it up.

PR #661, opened 2025-06-03
use a setting for the argument in time.sleep()

glbert-does:
and set it to 1 by default, but 0 for testing.


lakinwecker (approved):
I still want to find a way to remove these sleeps, but at least the tests aren't necessarily slowed down now.

PR #663, opened 2025-06-04
more tests for SeasonAdmin

glbert-does:
tests for SeasonAdmin in admin.py
this pr also contains #655.

edit: note that the test in the commit 1e98452 will fail without #678

### DIFF
--- a/heltour/tournament/tests/test_admin.py
+++ b/heltour/tournament/tests/test_admin.py
@@ -9,7 +9,7 @@ from django.http.response import HttpResponseForbidden
 from heltour.tournament.models import (Alternate, LonePlayerPairing, Player,
     Round, Season, SeasonPlayer, Team, TeamPairing, TeamPlayerPairing)
 from heltour.tournament.tests.testutils import (createCommonLeagueData,
-    get_season, get_player, get_round)
+    get_season, get_player, get_round, Shush)
 
 
 class AdminSearchTestCase(TestCase):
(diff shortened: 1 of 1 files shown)


heltour/tournament/tests/test_admin.py
```diff
@@ -1,50 +1,501 @@
...
+            self.path_s_changelist,
+            data={
+                "action": "simulate_tournament",
+                "_selected_action": get_season("lone").pk,
+            },
+            follow=True,
+        )
+        self.assertTrue(simulate.called)
+        self.assertTrue(message.called)
+        self.assertEqual(message.call_args.args[1], "Simulation complete.")
```
lakinwecker, line 86: Are the names of message and simulate mixed up here?

heltour/tournament/tests/test_admin.py
```diff
@@ -1,50 +1,501 @@
...
+        self.assertEqual(scalculate.call_count, 2)
+        self.assertTrue(message.called)
+        self.assertEqual(message.call_args.args[1], "Scores recalculated.")
+
+    @patch("django.contrib.admin.ModelAdmin.message_user")
+    @patch(
+        "heltour.tournament.admin.normalize_gamelink",
+        side_effect=[("incorrectlink1", False), ("mockedlink2", True)],
+    )
+    def test_verify(self, gamelink, message):
```
lakinwecker, line 129: Yeah, I'm confused by the order of these args, are they really in reverse order?

heltour/tournament/tests/test_admin.py
```diff
@@ -1,50 +1,501 @@
...
+            },
+            follow=True,
+        )
+        lpp2.refresh_from_db()
+        self.assertEqual(gamelink.call_count, 2)
+        self.assertTrue(message.call_count, 2)
+        self.assertEqual(
+            message.call_args_list[0][0][1], "1 bad gamelinks for Test Season."
+        )
+        self.assertEqual(message.call_args_list[1][0][1], "Data verified.")
```
lakinwecker, line 171: Surely there is a better API we can use to get the args we're after here?

lakinwecker (changes requested):
I just want answers to these questions, no need to fix them right now.

PR #664, opened 2025-06-04
more tests for models

glbert-does:
specifically and mostly ScheduledEvent, but also PlayerLateRegistration and ModRquest.


lakinwecker (approved):
LGTM

Issue #673, opened 2025-06-12
rethink registration validation


glbert-does:
this has been discussed for a while. i will do a summary of what *i* think should be changed in relation to this:

- [x] stop requesting the lichess profile of inactive players
- [x] instead request lichess profiles of players with pending registrations of active seasons once every couple of hours (24?)
- [x] when someone loads the registration page, make sure their profile has been requested from lichess in the past x hours
- [x] in a first step the reg validation task should not request any more information from lichess but use existing profile data only, to avoid hitting lichess with requests for individual players
- [x] in a second step there probably should not be a reg validation task, and the information about valid/invalid registrations should be fetched from the `player`-object as needed using an `@admin.display` decorator.
- [x] i believe specifically the previous point will be easier if `registration`s have a `player`-foreignkey (probably instead of their own lichess_username string).

whether all of this would work like i envision it remains to be seen.

PR #674, opened 2025-06-12
Registration updates

glbert-does:
get started on #673:
* stop fetching profiles of inactive players
* fetch profiles of registered players instead
* add a function "profile_update_after" to `Player`, that converts the lichess timestamp 'seenAt' to a datetime. could alternatively save the timestamp when we updated last.
* use that new function to decide whether a player profile needs updating when a player goes to the registration page
* if it does need updating, fetch the profile from lichess
* fix tests that now fail
* lint some of the files touched

### DIFF
--- a/heltour/tournament/models.py
+++ b/heltour/tournament/models.py
@@ -870,6 +870,17 @@ class Player(_BaseModel):
             self.profile = user_meta
         self.save()
 
+    def profile_update_after(self) -> datetime:
+        # lichess gives us "seenAt" in miliseconds as the last time the user was online
+        # thus, the profile was last updated *after* this seenAt.
+        if self.profile is None:
+            return None
+        seenAt = self.profile.get('seenAt')
+        if seenAt is None:
+            return None
+        else:
+            return datetime.utcfromtimestamp(seenAt / 1000)
+
     @classmethod
     def get_or_create(cls, lichess_username):
         player, _ = Player.objects.get_or_create(lichess_username__iexact=lichess_username,
--- a/heltour/tournament/models.py
+++ b/heltour/tournament/models.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 import logging
 import re
 from collections import defaultdict, namedtuple
-from datetime import date, timedelta
+from datetime import datetime, timedelta
 
 import reversion
 from ckeditor_uploader.fields import RichTextUploadingField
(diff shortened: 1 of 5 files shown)


heltour/tournament/models.py
```diff
@@ -826,6 +826,17 @@ def update_profile(self, user_meta):
...
+    def profile_update_after(self) -> datetime:
+        # lichess gives us "seenAt" in miliseconds as the last time the user was online
+        # thus, the profile was last updated *after* this seenAt.
+        if self.profile is None:
+            return None
+        seenAt = self.profile.get('seenAt')
+        if seenAt is None:
+            return None
+        else:
+            return datetime.utcfromtimestamp(seenAt / 1000)
```
lakinwecker, line 838:
```suggestion
        seenAt = (self.profile or {}) .get('seenAt')
        if seenAt is not None:
            return datetime.utcfromtimestamp(seenAt / 1000)
        return None
```

lakinwecker: I think the above is equivalent and is slightly easier to read, in my opinion.

heltour/tournament/tasks.py
```diff
@@ -22,89 +49,95 @@
...
-    return [p['lichess_username'] for p in users if p['lichess_username'].strip()]
+    return [p["lichess_username"] for p in users if p["lichess_username"].strip()]
 
 
 def just_username(qs: QuerySet[Player]) -> UsernamesQuerySet:
-    return qs \
-        .order_by('lichess_username') \
-        .values('lichess_username') \
-        .distinct()
+    return qs.order_by("lichess_username").values("lichess_username").distinct()
```
lakinwecker, line 65: This feels like something that might be nicely added to the QuerySetManager object for Player objects, rather than a method, but we don't have to do that right now.

heltour/tournament/tasks.py
```diff
@@ -22,89 +49,95 @@
...
-import sys, traceback
+import sys
+import traceback
 
 logger = get_task_logger(__name__)
 
 UsernamesQuerySet = ValuesQuerySet[Player, Dict[str, str]]
 
+
 def to_usernames(users: UsernamesQuerySet) -> List[str]:
```
lakinwecker, line 60: I think that using https://docs.djangoproject.com/en/5.2/ref/models/querysets/#values-list with `flat=True` is probably faster and more Djangonic (pythonic??) than this list.

heltour/tournament/tasks.py
```diff
@@ -22,89 +49,95 @@
...
 def active_player_usernames() -> List[str]:
     players_qs = Player.objects.all()
     active_qs = players_qs.filter(seasonplayer__season__is_completed=False)
     return to_usernames(just_username(active_qs))
 
-def not_updated_recently_usernames(active_usernames: List[str]) -> List[str]:
-    players_qs = Player.objects.all()
-    total_players = players_qs.count()
+
+def active_registrations_usernames(without_usernames: List[str]) -> List[str]:
```
lakinwecker, line 74: The name of this function feels misleading. This is is more of a
```python
def one_24th_of_registrations_needing_updates(...):
```

lakinwecker (commented):
Some minor initial lints

lakinwecker (approved):
LGTM

PR #675, opened 2025-06-16
remove old team generation algorithm

glbert-does:
to the best of my knowledge `createteams.py` has not been used in 6 or more years. in the unlikely event that we would need it again, i guess we can revert this commit.


lakinwecker (approved):
LGTM

PR #676, opened 2025-06-16
fix some ruff complaints

glbert-does:
in models.py:
* replace `condition == None` with `condition is None`
* fix typo `LonePlayerPlairing` => `LonePlayerPairing`

### DIFF
--- a/heltour/tournament/models.py
+++ b/heltour/tournament/models.py
@@ -1886,7 +1886,7 @@ class SeasonPlayer(_BaseModel):
     def has_scheduled_game_in_round(self, round):
         pairingModel = TeamPlayerPairing.objects.filter(team_pairing__round=round)
         if not self.season.league.is_team_league():
-            pairingModel = LonePlayerPlairing.objects.filter(round=round)
+            pairingModel = LonePlayerPairing.objects.filter(round=round)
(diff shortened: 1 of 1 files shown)


heltour/tournament/api.py
```diff
@@ -212,9 +231,9 @@ def _filter_pairings(pairings, player=None, white=None, black=None, scheduled=No
     if black is not None:
         pairings = pairings.filter(black__lichess_username__iexact=black) | pairings.filter(
             black__slack_user_id__iexact=black)
-    if scheduled == True:
+    if scheduled is True:
```
lakinwecker, line 234:
```suggestion
    if scheduled:
```

heltour/tournament/api.py
```diff
@@ -212,9 +231,9 @@ def _filter_pairings(pairings, player=None, white=None, black=None, scheduled=No
...
         pairings = pairings.exclude(result='', scheduled_time=None)
-    if scheduled == False:
+    if scheduled is False:
```
lakinwecker, line 236:
```suggestion
   else:
```

lakinwecker (changes requested):
Minor nits

PR #678, opened 2025-06-18
fix last_season_alternates if start_date is None

glbert-does:
if `Season.start_date` is not set, `last_season_alternates` will throw an error. fix that.

### DIFF
--- a/heltour/tournament/models.py
+++ b/heltour/tournament/models.py
@@ -286,7 +286,7 @@ class Season(_BaseModel):
         self.initial_is_completed = self.is_completed
 
     def last_season_alternates(self) -> set[Player]:
-        start_date = self.start_date if self.start_date is not None else timezone.now()
+        start_date = self.start_date or timezone.now()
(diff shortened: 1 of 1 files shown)


heltour/tournament/models.py
```diff
@@ -242,8 +243,9 @@ def __init__(self, *args, **kwargs):
         self.initial_start_date = self.start_date
         self.initial_is_completed = self.is_completed
 
-    def last_season_alternates(self):
-        last_season = Season.objects.filter(league=self.league, start_date__lt=self.start_date) \
+    def last_season_alternates(self) -> set[Player]:
+        start_date = self.start_date if self.start_date is not None else timezone.now()
```
lakinwecker, line 289: `start_date = self.start_date or timezone.now()`
I slightly prefer this, so long as ruff allows it.

lakinwecker (changes requested):
minor nit

lakinwecker (approved):
LGTM

PR #681, opened 2025-06-19
remove reg.classical_rating from workflows

### LINKED ISSUE #680: remove reg.classical_rating from workflows.py
https://github.com/Lichess4545/heltour/blob/3dc315ed3046adbcefd91d0c260badef9613bc84/heltour/tournament/workflows.py#L381
and https://github.com/Lichess4545/heltour/blob/3dc315ed3046adbcefd91d0c260badef9613bc84/heltour/tournament/workflows.py#L447
still assume that there exists a property Registration.classical_rating, which was removed.
fortunately, `self.player` should basically never be none in L381, and `player.rating is None` should almost never be true in L445, which is why this bug afaik has not really triggered any errors so far.

glbert-does:
closes #680.

### DIFF
--- a/heltour/tournament/workflows.py
+++ b/heltour/tournament/workflows.py
@@ -369,7 +369,7 @@ class ApproveRegistrationWorkflow():
         return self.reg.season
 
     @property
-    def default_ljp(self):
+    def default_ljp(self) -> float:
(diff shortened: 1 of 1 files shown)


lakinwecker (commented):
Why not use 1500 for rating in the rare cases we don't have it?

PR #683, opened 2025-06-21
send players a message after a game was scheduled

glbert-does:
which contains a link to the confirm button. message can be turned off via notification settings.
i don't think this should be a permanent solution though. once chesster supports confirming scheduled time directly, imo this can be removed.


lakinwecker (approved):
LGTM

PR #688, opened 2025-06-22
fix internal error for missing player

### LINKED ISSUE #687: ERROR (internal IP): Internal Server Error /admin/tournament/registration/
> AttributeError at /admin/tournament/registration/
> 'NoneType' object has no attribute 'rating_for'

> /heltour/tournament/models.py", line 1887, in rating
> return self.player().rating_for(league=self.season.league)

presumably some old registrations do not have a player object.

glbert-does:
closes #687.

i decided to simply create the player object if it is missing. this should also help to not trigger potential similar bugs elsewhere.

alternatively, we could check in the registration.rating property whether the player actually exists and return 0 as a rating otherwise.

### DIFF
--- a/heltour/tournament/models.py
+++ b/heltour/tournament/models.py
@@ -1880,7 +1880,7 @@ class Registration(_BaseModel):
             player__lichess_username__iexact=self.lichess_username).exclude(season=self.season)
 
     def player(self):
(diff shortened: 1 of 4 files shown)


lakinwecker (approved):
LGTM

PR #695, opened 2025-06-26
630 api tests

glbert-does:
* tests for api.py, so we can in the future avoid trivial bugs in `update_pairings` and friends.
* some small changes in api.py, so some functions are called with kwargs instead of args.
* lazily add typing to `get_round()` in testutils.py


glbert-does:
confirmed that these still work and rebased on current master.

PR #700, opened 2025-07-08
rethink registration validation

### LINKED ISSUE #673: rethink registration validation

glbert-does:
### motivation
currently we use the lichess api for information on [individual users](https://lichess.org/api#tag/Users/operation/apiUser) to periodically update information on registrations. the idea of this pr is to instead use the existing data of the user profile, which is already created when a user first logs in; and update it using the api for fetching information on [multiple users](https://lichess.org/api#tag/Users/operation/apiUsers), which we already use for active players anyway.
this also means that the periodic task would only check internal data to set binary variables (`validation_warning` and `validation_ok`), which seems unnecessary. instead these binary variables are calculated in the admin view.
for my own sanity i also remove the lichess_username field from the registration, and use a foreignkey for player instead. this is for consistency with how you would fetch a players name everywhere else, but it also simply makes more sense to me to store the name in one place only.

### intended outcomes
* validation task does not exist anymore
* registrations are categorised in valid etc. using existing information from the player object.
* the player object for registered players are updated every 24 hours, multiple at a time

### potential concerns
(apart from the usual)
* not sure how the admin view will do performance-wise, when it has to fetch data from player profiles
* does the migration from the field lichess_username to the foreignkey player work?
* do the updates of the player profile run as intended?
* is it problematic that regs don't get any automated comments anymore?
* is it problematic that regs can switch to valid and back to invalid again, e.g. if the RD is right around 110?

(i don't think so for the latter two, but second opinion would be nice.)

### changes
* use a player foreignkey in registrations instead of a lichess_username string [not strictly necessary for this pr, but introduced for consistency]
* remove regular validation task and use existing player information instead
* registrations are assigned "valid" status on the fly
* change the logic elsewhere to reflect the above changes
* note that mods can still request "validation" – but this only updates the player from lichess
* migrations are revertible, which probably makes them a bit slower. but i like to be able to revert the database on my dev instances.

closes #673.


heltour/tournament/forms.py
```diff
@@ -46,9 +47,11 @@ def __init__(self, *args, rules_url='', **kwargs):
         self.season = kwargs.pop('season')
         
         self.username = kwargs.pop('user')
-        
+
+        self.player, created = Player.objects.get_or_create(lichess_username=self.username)
```
lakinwecker, line 51: Won't this create and use a player with an empty username when the form is initially instantiated?

Pretty sure it will

lakinwecker: Also, why are we using  a `username` field in this case. Can't we enforce a Player object to be in scope because you've already signed in?

heltour/tournament/models.py
```diff
@@ -1899,14 +1911,19 @@ def was_rejected(cls, user, season):
 
     @classmethod
     def get_latest_registration(cls, user, season):
-        return (cls.objects
-                .filter(lichess_username__iexact=user.username, season=season)
-                .order_by('-date_created')
-                .first())
+        return (
+            cls.objects.filter(
+                player__lichess_username__iexact=user.username, season=season
```
lakinwecker, line 1916: There must be a way we can get around using username here, right?

heltour/tournament/models.py
```diff
@@ -1899,14 +1911,19 @@ def was_rejected(cls, user, season):
...
+            )
+            .order_by("-date_created")
+            .first()
+        )
 
     @classmethod
     def is_registered(cls, user, season):
-        return cls.objects.filter(lichess_username__iexact=user.username, season=season).exists()
+        return cls.objects.filter(
+            player__lichess_username__iexact=user.username, season=season
```
lakinwecker, line 1925: Same here?

heltour/tournament/notify.py
```diff
@@ -135,7 +135,7 @@ def player_account_status_changed(instance, old_value, new_value, **kwargs):
         return
 
     season_players = instance.seasonplayer_set.select_related('season__league').nocache()
-    pending_regs = Registration.objects.filter(lichess_username__iexact=instance.lichess_username,
+    pending_regs = Registration.objects.filter(player__lichess_username__iexact=instance.lichess_username,
```
lakinwecker, line 138: Same here.

heltour/tournament/tasks.py
```diff
@@ -69,25 +68,28 @@ def active_player_usernames() -> List[str]:
...
+    active_regs = (
         Registration.objects.filter(
-            status__exact="pending", season__registration_open=True
-        ).exclude(lichess_username__in=without_usernames)
+            status__exact="pending",
+            season__registration_open=True,
+            player__date_modified__lte=_24_hours,
+        )
+        .exclude(player__lichess_username__in=without_usernames)
+        .values_list("player")
```
lakinwecker, line 79: Do you need flat=True here? I don't remember

lakinwecker (changes requested):
I think the creation of a Player with an empty username in the registrationform is my primary concern right. Overall the approach seems fine, but I think that we need to be careful with this one, like you said.

I feel that it might be worth doing some work to improve our ability to get the "Player" objects from the user objects to avoid having to pass around and use usernames

heltour/tournament/models.py
```diff
@@ -1864,27 +1864,39 @@ class Registration(_BaseModel):
...
-        return Player.get_or_create(lichess_username=self.lichess_username)
+    @property
+    def lichess_username(self):
+        return self.player.lichess_username
 
     @property
     def rating(self):
-        return self.player().rating_for(league=self.season.league)
+        return self.player.rating_for(league=self.season.league)
+
```
lakinwecker, line 1887: We do not need to change this to be able to merge and deploy, but I think we could define an Enum for the validation status, and then have a single validation status function here, which returns the appropriate status, and use the python `match` thing to deal with it when we use it.

lakinwecker (approved):
Yeah, I say we just YOLO this live

PR #711, opened 2025-07-31
specify lichess4545 in the user-agent

glbert-does:
be more specific about who we are in the user-agent.
not sure if the heltour version is already specified anywhere, i put it into the settings. though i have found some information that it could go into `manage.py` as `__version__`, but it seems to me it would be more difficult to find there.

edit: also let ruff re-sort all the imports in `heltour/api_worker/`

edit 2: not sure if a user-agent should be sent in the websocket connection in `worker.py` as well.

### DIFF
--- a/heltour/api_worker/urls.py
+++ b/heltour/api_worker/urls.py
@@ -1,4 +1,5 @@
 from django.urls import path
+
 from heltour.api_worker import views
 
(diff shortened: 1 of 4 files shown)


lakinwecker (approved):
LGTM

PR #712, opened 2025-08-02
Sync litour with heltour

lakinwecker:


pyproject.toml
```diff
@@ -1,15 +1,14 @@
 [tool.poetry]
 name = "heltour"
 version = "1.0.1"
-description = "A [hel]per for [tour]naments.  Intended to run swiss based tournaments, but is actually the source for the lichess4545.com leagues."
+description = "A [hel]per for [tour]naments.  Intended to run swiss based tournaments, but is actually the source for the lots.lichess.ca leagues."
```
glbert-does, line 4: maybe we should consider making this more general and just write something like "source for leagues run on lichess". "swiss based" might become outdated too, if at some point we branch into knockouts.

PR #716, opened 2025-08-03
add missing receivers for broadcast signals

glbert-does:
also, get started on a first test for the admin actions.


lakinwecker (approved):
LGTM

PR #717, opened 2025-08-04
log post-data

glbert-does:
adds a simple function to truncate values in a string + censor them if they are tokens.
then logs post-data in api calls.


heltour/api_worker/views.py
```diff
@@ -18,13 +19,24 @@
 # Optimize for typical rate limits: 1, 2, 5, 10 minutes
 retry_wait_times = {0: 60, 1: 60, 2: 180, 3: 300, 4: 600}
 
+
+def _trunc_and_clean(input: dict) -> dict:
+    output = {}
+    for key, value in input.items():
+        # censoring tokens
+        cleanvalue = re.sub(r"(li[opu]_)([A-Za-z0-9]+)", r"\1***", str(value))
```
lakinwecker: I wouldn't mind seeing some unit tests for this regex so that we can ensure it's doing what we want.

Wrap it in a function. call it here and make some tests

lakinwecker (changes requested):
Want some tests.

PR #718, opened 2025-08-06
broadcasts: fix several bugs

### LINKED ISSUE #714: rename "first_board" in broadcasts to something sensible
for example `board_range_start`

### LINKED ISSUE #722: broadcast: pairing may be added to two sub-broadcasts
happened at least in team s45r1, where pairings #97 and #98 to both subbroadcasts, while they should only be in the lower subbroadcast.

edit:
https://github.com/Lichess4545/heltour/blob/1c3f5585434e0d98ce791e85890d7d14501e9582/heltour/tournament/tasks.py#L599 should not use the max-parameter here. maybe not for lw-leagues either, but there it is less of a problem.

glbert-does:
* remove useless titles from grouping, since lichess ignores them anyway.
* fetch actual last board for sub-broadcast titles, instead of just assuming it is the lichess max.
* send broadcast update signal in a loop independent from autostarting, to ensure we are updating non-autostarted games even when there are no autostarted games.
* use the last board getter when deciding on which pairings are in which broadcast (closes #722)
* rename `first_board` to `board_range_start` and `last_board` to `board_range_end`. (closes #714)


heltour/tournament/models.py
```diff
@@ -757,13 +757,25 @@ def get_league(self):
     def is_team_league(self):
         return self.season.league.is_team_league()
 
-    def get_broadcast_id(self, first_board: int = 1) -> str:
-        return self.season.get_broadcast_id(first_board=first_board)
+    def get_broadcast_round(self, board_range_start: int = 1) -> Broadcast|None:
+        bc = Broadcast.objects.filter(
```
lakinwecker, line 761: What happens if there is no broadcast here? Do we get an error?

heltour/tournament/models.py
```diff
@@ -757,13 +757,25 @@ def get_league(self):
...
-    def get_broadcast_round_id(self, first_board: int = 1) -> str:
+
+    def get_broadcast_id(self, board_range_start: int = 1) -> str:
+        return self.season.get_broadcast_id(board_range_start=board_range_start)
+
+    def get_broadcast_round_id(self, board_range_start: int = 1) -> str:
         if not self.season.get_broadcast_id():
             return ""
-        bc = Broadcast.objects.get(season=self.season, first_board=first_board)
+        bc = Broadcast.objects.filter(
```
lakinwecker, line 777: Same here. Not quite sure what the best approach is.

lakinwecker (changes requested):
A couple of minor comments, not sure any changes are needed, I just wouldn't mind knowing (either via a code comment, or comment on this PR) why this approach is what we're using and what the expected behavior is?

heltour/tournament/models.py
glbert-does: if there is no broadcast, then `.first()` on the empty `QuerySet` should return `None`. filtering the next QuerySet by `broadcast=None` should return another empty `QuerySet`, and the `first()` should make the whole function return None.
i will add a comment for it.

glbert-does: the type hint was incorrect though.

lakinwecker (approved):
LGTM

PR #721, opened 2025-08-08
put the ordering where it belongs

glbert-does:
ordering the registrations is completely useless when we return the players.


lakinwecker (approved):
LGTM

PR #728, opened 2025-08-19
fix pre-round report for teams

### LINKED ISSUE #725: Pre-round report shows no games without results when a game is still to be played
Before creating new pairings for round 2 of season 45, the pre-round report showed that there were no pairings without results.

Image

However, one game was still to be played (and later was ongoing).

Image

Expected behavior is that if a game is still to be played or is ongoing (i.e. there's no result yet) that this shows up in the Pre-round Report as a game without a result.

glbert-does:
fixes #725


glbert-does:
looks like what this fixes is not the source of the bug in 725, unfortunately.

PR #747, opened 2026-03-17
fix red cards on rosters page

NirakuHakamura:
Fix roster status precedence so red-carded players are displayed with the red-card marker instead of only as unavailable. This ordering was introduced in 9577693e and the bug became effective 100 % of the time when 694c7d44 started marking red-carded players unavailable automatically.

### DIFF
--- a/heltour/tournament/templates/tournament/team_rosters.html
+++ b/heltour/tournament/templates/tournament/team_rosters.html
@@ -50,7 +50,11 @@
                                             {% for board_number, board in team.boards %}
                                                 <td class="cell-player{% if board.is_captain %} captain{% endif %}"
                                                     colspan="2">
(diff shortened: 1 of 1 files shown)


glbert-does:
still not sure what the preferred order of preference for display should be here. as far as i am concerned there is an argument for the current order, but i dunno.

lakinwecker (approved):

lakinwecker:
From a technical perspective I'm happy with the change. I'd prefer for some amount of the mods to make a decision and tell me whether to merge this or not.

glbert-does:
i thought about it for a bit in the meantime, and now i personally think the change is in fact preferable.
i will quickly ask the team mods though.

glbert-does:
4/4 so far want the pr merged.

PR #758, opened 2026-07-12
feat: use dagger.io for CI definition in typescript.

lakinwecker:


lakinwecker:
Not pursuing this now, it's just annoying enough to run in github CI. when I asked about how to do caching better on github CI the answer I got was: self-hosted runners.  And when I asked about instructions for how to do caching on self hosted runners, they linked me to how to run them on kubernetes. Which, while a reasonable answer for many situations is not a particularly great answer for us.

Issue #628, opened 2025-05-07
reducing reversion usage


glbert-does:
we would like to remove reversion for most objects, and leave it mostly for alternate-related changes and documents.
for example, reversion for many of the api-call related changes are redundant since these are triggered by commands on our communication platform, and nobody is going to revert those changes using reversion, because they are trivially reverted by edits of the objects.

PR #762, opened 2026-07-31
fix unresponsive alternate queue position

### LINKED ISSUE #761: Unresponsive alternates stay at the top of the queue
# Unresponsive alternates stay at the top of the queue

*Maintainers/admins: please consider giving this issue the P1 label.*

---

## TL;DR

Rule X.D:

> *"Any player not responding within 24 hours of an offer will be moved to the bottom
> of the list as an alternate and be marked as unresponsive."*

In practice this almost never happens: the deadline is only checked at rare
"snapshot" moments, and every re-contact (a fresh offer to the same alternate once a
new spot opens) restarts the clock. On top of that, the live configuration
apparently overrides the deadline to **72 hours** — the offer DMs literally tell
alternates they have three days — in direct contradiction of the rules (see the
warning below; this part the mods can fix today, no code change needed). The result
is the well-known phenomenon of the same names sitting at the top of the alt lists
round after round, clogging the pipe. The ones paying for it are the alternates who
actually want to play and wait 24h+ for an offer that should have reached them much
sooner — and just as much the player stuck without an opponent and the team watching
the search grind: waits regularly stretch well past a full day, and alt spots keep
going unfilled before the round ends as a direct consequence of this behavior.

## The cast

Four recurring roles, in descending order of pipe-clogging:

- **The rogue alt ("RogueAlt")** — the worst offender: signed up as an alternate, has
  never played a single game, never responds to anything. Since queue priority is
  based on the registration / "made alternate" date, and only playing a game (or a
  demotion that never comes, see below) moves you back, RogueAlt owns a **permanent
  seat at the very top of the queue**. Every new vacancy pings RogueAlt first.
  Nothing ever happens.
- **The sleepy alt ("SleepyAlt")** — did play a game at some point, which moved them
  down for a while, but has since gone into hibernation and stopped responding. As
  others play and rotate to the back, the sleeper drifts back up the list — and from
  there behaves exactly like RogueAlt until the end of the season. Sleepers rarely
  come alone.
- **The dutiful alt ("DutifulAlt")** — can't always play, but always answers:
  declines promptly — at the latest within the 24 hours the rules ask for — or
  accepts when the schedule allows. A prompt answer costs the pipe nothing — after a
  decline the search contacts the next alternate immediately instead of waiting out
  the 8h interval. Yet the queue treats the dutiful exactly the same as the alts who
  never answer at all. This role also comes in the flavor of the ReallyDutifulAlt,
  who sets themselves unavailable in advance. However, unavailable alternates have
  no influence on the filling of open spots and are thus out of scope of this issue.
- **The keen alt ("KeenAlt")** — is glued to Slack and would drop everything to
  play. KeenAlt's tragedy: sitting below the sleepers, every offer arrives a working
  day late — or, with enough clog upstream, not at all before the alt search for the
  current round closes.

And two spectators who did nothing wrong and suffer the most:

- **The stranded player ("StrandedPlayer")** — has an opponent-shaped hole in their
  pairing and refreshes Slack instead of scheduling their game.
- **The waiting team ("WaitingTeam")** — is short a board and watches a long column
  of alts on "Waiting" who would happily play, kept out of reach by the few clogging
  the pipe.

## Current behavior

The broken mechanics, in four points:

1. Offers go out one `contact_interval` apart (8h in-round, 12h before the round
   starts). An alternate who ignores their offer is simply **skipped** — while a
   search is running, no deadline is ever checked.
2. The deadline is only evaluated at two **snapshot moments**: when the board's open
   spots hit zero, or when the next round's search starts. Anyone whose *latest*
   contact is younger than `unresponsive_interval` at that instant is quietly reset
   to "waiting" — unpunished, no matter how long they have been ignoring offers in
   total.
3. When the next spot opens, the queue is rebuilt and the same sleepers are
   contacted **again — which overwrites `last_contact_date` and restarts the
   deadline clock.** Silence therefore never accumulates into "24h unresponsive" at
   a moment where it would count; the clog carries over from round to round.
4. All the while the ignored alternate's accept link **stays valid**: wake up at
   hour 60, click accept, and the spot is yours — ahead of every KeenAlt patiently
   holding the line below. Rule X.D promises the bottom of the list; the code grants
   a right of first refusal.

Net effect: RogueAlt and the sleepers tax every alternate search with one contact
interval apiece, round after round; demotion fires only by coincidence of timing;
and as a bonus, the queue displayed on the website does not match the order in which
alternates were actually contacted (the "strange listing order" players keep
noticing).

## Expected behavior

Rule X.D (quoted at the top) is actually quite clear: an alternate who has not
responded within 24 hours **of an offer** should be marked unresponsive and moved to
the bottom of the queue. RogueAlt and the sleepers sink
to the bottom, offers reach responsive alternates much earlier and spots fill in a fraction of the
time.

> ⚠️ **The 72-hour configuration contradicts the rules and should be reverted
> immediately — independent of any code fix.** The offer DMs currently tell
> alternates *"please click one of the following links within 72 hours"*, which
> means the live `unresponsive_interval` has been raised from the default 24h to
> 72h. This is doubly harmful: it officially invites every alternate to sit on an
> offer for three days, and it triples the (already almost-never-enforced) deadline
> after which the queue could clean itself. A further argument against the 72h: why
> should alternates have 72h when full-time players are given only 24h to respond
> to pairing messages? **Mods: please set `unresponsive_interval`
> back to 24h in the live config to match rule X.D — this
> takes a minute and helps today, even before any of the bugs are fixed. It also
> re-establishes alignment with the rules.**

## Suggested Fix

- **Check the deadline continuously:** on every manager tick, mark any `contacted`
  alternate with `now - last_contact_date > unresponsive_interval` as unresponsive
  (and demote), independent of spot churn.
- **Demotion should change the queue order, not the ability to accept.** Today the
  two are coupled: marking an alternate unresponsive kills their response links
  (`alternate_accepted` requires status `contacted`). A fix should decouple them —
  demote at the deadline, but keep the accept/decline links working while a search
  on that board is still active. On boards where the pool has run dry, the silent
  contacted alternates are the only hope left; without this, enforcing rule X.D
  would make thin boards *harder* to fill, trading games played for queue justice.

NirakuHakamura:
## Summary

Fixes #761.

On every manager tick, contacted alternates whose latest offer is older than
`unresponsive_interval` are marked unresponsive and moved to the bottom of the
queue. This deadline check is no longer limited to round transitions or the
point at which a board's open spots are filled. Their existing accept and
decline links remain usable while a spot is still open.

The implementation and the regression tests are separated into two commits for
easier review: the first commit is the complete fix, the second contains tests
only and can be dropped if a smaller test scope is preferred.

## Previous behavior

The deadline was only enforced at two snapshot moments: when all of a board's
open spots were filled, and when the next round's search started. In addition,
every re-contact overwrote `last_contact_date` and restarted the clock.
Alternates who never respond could therefore sit at the top of the queue for a
whole season, taxing every search with one `contact_interval` each — the
"same names at the top of the alt lists" phenomenon described in #761. When
the demotion did fire, it also killed the alternate's response links
(`alternate_accepted` required status `contacted`).

## New behavior

1. The alternates manager checks overdue contacted alternates on every tick
   (every 2 minutes via Celery Beat).
2. Overdue alternates are marked unresponsive; the status change is re-checked
   under a database row lock so a concurrent accept/decline is never
   overwritten.
3. The marking moves the alternate to the bottom of the queue via the existing
   `priority_date_override` mechanism.
4. A Slack DM explains the demotion (new `alternate_unresponsive` signal).
5. The accept and decline links remain usable: both the manager functions and
   the views now treat `unresponsive` like `contacted`, so on boards where the
   pool has run dry a late alternate can still fill an open spot.

Deliberately unchanged: the deadline stays **per offer** (matching the wording
of rule X.D — a legitimately re-contacted alternate gets a fresh clock for the
fresh offer), and the round-rollover reset and the lifetime of the demotion are
untouched.

### Unresponsive notification

No staged Slack screenshot — this is the exact DM text (see
`notify.alternate_unresponsive`), with the interval formatted from the league
setting:

> Since you didn't respond to the alternate offer within 24 hours, you've been
> moved to the bottom of the alternates list. You can still accept the offer
> using the links you received, as long as the spot is open.

### Late acceptance remains available

The accept page for an alternate that has been marked unresponsive while the
spot is still open (simulated league, fictional players):

Late acceptance remains available

### Queue position before and after

A simulated scenario reproducing the behavior described in #761, replayed
against master and against this branch with identical configuration on both
sides (`unresponsive_interval = 24h`, the code default) — the only variable is
this code change. A spot opens Monday
noon and nobody accepts for over two days. On master **no alternate is ever
marked**: as long as the board has an open spot, no deadline check runs, so by
Wednesday RogueAlt has been sitting on the offer for 49 hours and even
SleepyAlt3 is past the deadline — all still plain "Contacted". With this fix,
each of them is marked unresponsive and demoted at the first tick past their
own 24h deadline, mid-search, while the dutiful alternates who answered stay
untouched:

#### Before

Queue before the fix

#### After

Queue after the fix

Note that the live configuration currently appears to be 72h (see the warning
in #761) — at that setting, master marks nobody at all; this PR makes whatever
interval is configured actually enforceable.

## Testing

- Deadline boundary just before and just after the configured interval (ages
  derived from the setting, not hardcoded)
- Queue demotion (demoted alternate sorts to the bottom)
- Manager tick integration
- Notification content and exactly-once delivery
- Accept and decline for unresponsive alternates, at the manager level and
  end-to-end through the logged-in Django views (including the pairing swap on
  accept)

Results:

```text
python manage.py test heltour.tournament.tests.test_alternates_manager
12 tests passed

python manage.py test heltour.tournament.tests
139 tests passed
```


heltour/tournament/alternates_manager.py
```diff
@@ -57,10 +58,40 @@ def tick(season):
...
+                                                  last_contact_date__lt=deadline) \
+        .select_related('season_player__player').nocache()
+    for alt in overdue_alternates:
+        with transaction.atomic():
+            # Re-check under lock so a concurrent accept/decline isn't overwritten
+            locked_alt = Alternate.objects.select_for_update().get(pk=alt.pk)
+            if locked_alt.status != 'contacted' or locked_alt.last_contact_date is None \
+                    or locked_alt.last_contact_date >= deadline:
+                continue
+            with reversion.create_revision():
```
glbert-does: note that we have an issue for reducing reversion usage: #628
i doubt any mod would ever use reversion for managing alternates, so this is probably rather useless.

NirakuHakamura: I hadn't seen #628.

Reading it now, I'm not sure I see the conflict. The issue says reversion should go for most objects but be kept "mostly for alternate-related changes and documents", which is exactly what this block is — and it follows the rest of alternates_manager.py, where every other status transition opens a revision.

Even if #628 ends up covering alternates too, that sweep will remove this block along with the dozen others in the same file — so keeping it costs nothing at that point. Dropping just this one now is the option that leaves the file inconsistent in the meantime, and #628 has been open for a while, so I'd rather not bet on the timing.

That said, it's your call — if you'd rather not add new reversion usage while #628 is open, say the word and I'll take it out, but it would leave the inconsistency until #628 is implemented.

PR #764, opened 2026-08-01
feat: prioritize single-open alternate searches

### LINKED ISSUE #534: For alternate searches, deprioritise matches that would be between two alternates.
Only for when both availabilities are known before the round starts. This way, more regular players get to play their match, and if any matches aren't played it prioritises those where both regular players were unavailable.

NirakuHakamura:
## Summary
Fixes #534

When an alternate accepts, searches for pairings with only one open spot are now prioritized over pairings where both teams need an alternate.
The existing search order remains unchanged, i.e. within each priority group, the oldest search is still selected first.

### Before alternate accepts offer to play
before

### After alternate accepts offer to play
after

## Testing
- Added coverage for the A-B-A search sequence described in the issue.
- Ran `python manage.py test heltour.tournament.tests.test_alternates_manager`.
- Verified the assignment manually in the browser with and without the change.


NirakuHakamura:
Holding this one back until after #762 has been merged.

Issue #305, opened 2019-04-17
Dont make a DM Group when you assign someone available in a middle of the round


Moha369:
When assign someone unavailable before the round begin chesster will not make a DM Group
But if you assign the player available again in the middle of the round chesster will not make a DM group

hopffgam:
Assigned medium priority. The behavior described in this issue leads to confusion among players but it happens comparably rarely as it requires a player to go from unavailable to available after round start.

A possible solution would be to simply always trigger a pairing message if a player on a pairing goes available.

glbert-does:
this is the same as #469, right? i think it can be closed as a duplicate.

PR #765, opened 2026-08-02
fix: pairing notification after availability change

### LINKED ISSUE #469: Pairing messages not send when player was unavailable at round start
In the following situation no pairing message is sent: 
1. Team player A gets unavailable before round start.
2. No alternate is found until round start
3. At round start the opponent B  of player A gets a message from chesster that we are still searching for an alternate
4. Before an alternate is found, player A decides to switch to available and play the game. The alternate search is now stopped but A and B do not get a pairing message.

Expected: there should be a pairing message in the above scenario when player A goes available in step 4.

An additional consequence of this issue is that if the game does not get scheduled within 24 hours we assume that both players are unresponsive which is wrong as without pairing message channel we cannot even check the responsiveness.

NirakuHakamura:
## Summary

Fixes #469, #305

When a team player was unavailable at round start, the normal pairing notification was intentionally skipped while the alternates manager searched for a replacement. If the original player later became available before an alternate was assigned, the search stopped, but neither player received the missing pairing message or Slack group DM.

This change records whether the round-start pairing notification was sent or explicitly skipped. On the next alternates-manager tick, a search that is no longer needed is cancelled and the missing pairing notification is sent exactly once when both players are available and the pairings are published.

It also covers the scenario reported in #305, which was closed as a duplicate of this issue.

## Important: database migration required

> [!IMPORTANT]
> This PR adds the nullable database column
> `PlayerPairing.round_start_notification_sent` through migration
> `0015_playerpairing_round_start_notification_sent.py`.
>
> **The migration must be applied before the new web or Celery code is started.**
> Running the new code against an unmigrated database can fail with a
> `ProgrammingError` because Django ORM queries expect the new column to exist.
>
> Safe deployment order:
>
> 1. Apply the migration with `python manage.py migrate`.
> 2. Start or restart the updated web and Celery services.
>
> The migration is additive and nullable, performs no data rewrite or backfill,
> and can safely be applied while the previous application version is still
> running.

## Behavior

- A pairing notification skipped because of player unavailability is marked explicitly.
- When the original player returns and no alternate has been assigned, the open search is cancelled on the next regular alternates-manager tick.
- If both players are available and pairings are published, the missing pairing notification is sent exactly once.
- If pairings are not yet published, the search is cancelled and the normal round-start notification remains responsible for messaging later.
- If both players were unavailable, no pairing message is sent until the pairing has two available players.
- Completed searches and already assigned alternates are left untouched.
- A cancelled search can be restarted if the original player becomes unavailable again.
- Pairings that already received their round-start notification never receive a duplicate.

## Implementation notes

The nullable status has three meanings:

- `NULL`: existing or not yet classified pairing
- `False`: round-start notification was skipped because an original player was unavailable
- `True`: the normal, late, or alternate-assignment pairing notification was sent

The recovery path uses the existing `notify_players_late_pairing` signal, so it follows the same notification code as other late pairings.

## Testing

```text
HELTOUR_ENV=STAGING python manage.py test heltour.tournament.tests
131 tests passed
```

Additional checks:

- `python manage.py makemigrations --check --dry-run`: no model changes missing from migrations
- `python manage.py check`: no system-check issues
- `git diff --check`: clean


heltour/tournament/alternates_manager.py
```diff
@@ -103,6 +103,44 @@ def reset_alternate_search(season, round_, setting):
...
+        if pairing.white.is_available_for(round_) and pairing.black.is_available_for(round_):
+            pairings_to_notify[pairing.pk] = pairing
+
+    for pairing in pairings_to_notify.values():
+        signals.notify_players_late_pairing.send(sender=do_alternate_search, round_=round_,
+                                                 pairing=pairing)
+        time.sleep(SLEEP_UNIT)
+
+    for search in resolved_searches:
+        with reversion.create_revision():
```
glbert-does: similar to the other pr, i don't think reversion is very useful here.

NirakuHakamura: This one isn't new reversion usage, though — it's the same four lines master already has in `reset_alternate_search`:

https://github.com/Lichess4545/heltour/blob/6000ce553094e62f929e61d4784a4eb1f1fa9477/heltour/tournament/alternates_manager.py#L94-L98

Same model, same `cancelled` transition, same revision comment — I'm only reaching it from a new path (on a tick, once the search is moot, instead of only during the next-round reset). Dropping it here would mean the same status change is recorded during the next-round reset but not when it happens mid-round.

Happy to take it out if you'd rather. If so, I'd offer to strip the existing one in `reset_alternate_search` too, so both paths match — as a separate PR. Or we leave both for the #628 cleanup, if you'd prefer that.

